Issue #3, opened 2022-03-06
Convert ARX forecaster to the correct format


dajmcdon:
The proper forecaster signature should be:

```
prob_arx <- function(x, y, geo_value, time_value, args = prob_arx_args())
```

All args not in the `epi_` data frame should be in an args list.

brookslogan:
I think we may want to consider instead
```
function(df, args = prob_arx_args())
```
or something mirroring `fable`, whatever its interface is.

As:
- More commonly than not, users will probably will want (x,y,geo_value,time_value) in a data frame format to pass to a fitter function like `lm`, `rq`, etc.  If they are given `df` as an input, this is less trouble.
- The code to call a forecaster / list of forecasters seems easier:
```
fetch_latest_data_from_my_favorite_api() %>% prob_arx(my_args)
as_of(archive, forecast_date) %>% prob_arx(my_args)
snapshot = as_of(archive, forecast_date)
lapply(forecasters_with_args, function(forecaster_with_args) {
  forecaster_with_args$forecaster(snapshot, forecaster_with_args$args)
})
```
as opposed to something with `do.call` or `tidy_eval` or things along those lines.  Although I'm not sure if taking `df` as an argument would make things easier or harder when using `epi[x]_slide`; I forget if slider expects functions to take in the columns or the df.

dajmcdon:
Addressed in #10

Issue #8, opened 2022-03-09
What signature should a forecaster object have?


dshemetov:
Making this thread to address the broader question opened in #3 about the function signature of a forecaster object. cc @dajmcdon @brookslogan @ryantibs 

Current proposals:
1. A combination of features, response variables, a few index args exposed, and the rest bundled into a control arg
```r
prob_arx <- function(x, y, geo_value, time_value, args = prob_arx_args())
```

2. A plain data interface and the rest of the arguments bundled into a control arg (see Logan's full comment [here](https://github.com/cmu-delphi/epipredict/issues/3#issuecomment-1061193088))
```r
function(df, args = prob_arx_args())
```

Two dimensions of variation here: 
- do forecasters need to be aware of the index variables, if they are already being given axis-stripped training/testing matrices?
- does everything except data go into the control arg? what columns does the forecaster expect from `df`?

One possibility (setting aside the question of whether indexing variables need to be arguments) is that the forecaster signature in (1) could be a function inside the forecaster in signature (2). (2) would do additional parsing of the dataframe `df` and produce training/testing matrices `x`, `y` and hand them off to something that looks like (1).

A question to pump our design intuition: what do we want a forecaster call pattern to look like? For example, following along with the [vignettes](https://cmu-delphi.github.io/epiprocess/articles/archive.html) in epiprocess, we can imagine having a call that looks something like this:

```r
library(covidcast)
library(epiprocess)
library(data.table)
library(dplyr)

forecast_date <- "2021-12-01"
dv <- covidcast_signal("doctor-visits",
                       "smoothed_adj_cli",
                       start_day = "2020-06-01",
                       end_day = "2021-12-01",
                       issues = c("2020-06-01", "2021-12-01"),
                       geo_type = "state",
                       geo_values = c("ca", "fl")) %>% 
           select(geo_value, time_value, version = issue, percent_cli = value) %>%
           as_epi_df()

forecasts <- dv %>% group_by(geo_value) %>%
                     map(some_forecaster, forecast_date = forecast_date, args = prob_arx_args())
```

This assumes `some_forecaster` to have a similar signature as when we are using `evalcast`, namely `forecaster(df_list, forecast_date)` (I'm not entirely sure if I can pipe the result of the group_by into map like that, but let's suppose that's correct).

That is one possibility. Another possibility is to have an `evalcast::get_predictions`-like function
```r
[...]

PR #10, opened 2022-03-15
Refactor arx

dajmcdon:
Some basic checks pass, but there are a number of questions throughout.


dajmcdon:
Here's the plot that @ryantibs made in the `epiprocess` [vignette](https://cmu-delphi.github.io/epiprocess/articles/advanced.html#version-aware-forecasting-revisited-1). Looks similar, though not exactly the same.

Rplot

dajmcdon:
Closes #3 and #8

PR #19, opened 2022-05-05
Add support for knn enhanced ar(x) forecastors

lzlzlizi:
Hi, @dajmcdon. Those commits add the support for KNN enhanced iterative, dynamic iterative, and direct methods. The direct method support the `arx` mode while the two iterative methods only support the `ar` mode. 

A few comments:
1.  The search-train-predict-residual procedures are tangled together because we need the `new_data` first to start.  Although we can technically split some of them apart, splitting will induce non-trivial space or computation redundancy.
2. `arx_forecastor.R` was accidentally changed when I apply auto formating for the scripts. Those changes can be revoked if needed.
3. KNN searching is supported by package `tensr`. Do we need to manually add it to dependencies?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -40,6 +40,7 @@ Suggests:
     lubridate,
     RcppRoll,
     rmarkdown,
+    tensr,
     testthat (>= 3.0.0)
 Config/testthat/edition: 3
 URL: https://github.com/cmu-delphi/epipredict/,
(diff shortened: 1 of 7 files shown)


lzlzlizi:
Sorry for bombarding you with the emails of R-CMD-check failures. Didn't realize it will send emails to everyone. I will temporarily close the PR until I fixed the check failure issue.

Issue #22, opened 2022-05-20
Unify the `epi_lag` and `epi_ahead` functions


dajmcdon:
These are basically the same. Try to remove as much extraneous repetition as possible (say by dispatching to common underlying functions).

kenmawer:
Currently on my fork, I'm having an error here. I wilk continue work on it tomorrow.
![example-recipe errors](https://user-images.githubusercontent.com/64671957/170392507-a913407b-469e-40a8-9dc6-9c0785c9164b.png)
e

dajmcdon:
I cannot reproduce your error locally. Can you describe how you are running the script? What R/package versions you're using?

kenmawer:
> I cannot reproduce your error locally. Can you describe how you are running the script? What R/package versions you're using?

On the example recipe, I run this with Ctrl+Shift+R (can also do Ctrl+A then Ctrl+R).

R version 4.1.1

tidyverse 1.3.1

covidcast 0.4.2

delphi.epidata 1.0.0

epiprocess 1.0.0

tidymodels 0.2.0

Note that I am getting the error with the code shown on my kenmawer/epipredict. I haven't tried it out with cmu-delphi/epipredict yet despite it differing from mine due to the inclusion of epi_shift.

dajmcdon:
Have you loaded `epipredict`? (with `devtools::load_all()`) Note: the function `epi_shift()` is never called by that example script. (It is used in other, unrelated parts of the package?

kenmawer:
> Have you loaded `epipredict`? (with `devtools::load_all()`) Note: the function `epi_shift()` is never called by that example script. (It is used in other, unrelated parts of the package.)

Yes, I have. I think we should discuss this issue in detail as I followed all your steps but I still get this error, implying I may have done something else wrong.

kenmawer:
Just a note: I still need to work on improving documentation for this, as the documentation is not quite right.

dajmcdon:
Closed by #77

PR #31, opened 2022-05-25
Km issue #22: unite

kenmawer:


kenmawer:
I did this by accident. Please ignore it.

PR #34, opened 2022-06-01
Completed unification of ahead and lag, as well as improving print

kenmawer:


kenmawer:
@dajmcdon I cannot assign you for review, so I am leaving this here.

dajmcdon:
That's because it's a "Draft Pull Request". If it's ready for review (tests are failing), click the "Ready for Review" button.

R/step_naomit2.R
```diff
@@ -0,0 +1,13 @@
+#' Special NA omit step that does two steps in one
+#'
+#' @param x Recipe to be used for omission steps
+#'
+#' @return Omits NA's from both predictors and outcomes and training time;
+#' however, only omits predictors at prediction time
+#' @export
+
+step_naomit2 <- function(x) {
```
ChloeYou, line 9: I'm not sure if it's common to add a number in a function name? Perhaps something like step_naomit_all()?

R/step_naomit2.R
```diff
@@ -0,0 +1,13 @@
+#' Special NA omit step that does two steps in one
+#'
+#' @param x Recipe to be used for omission steps
+#'
+#' @return Omits NA's from both predictors and outcomes and training time;
+#' however, only omits predictors at prediction time
+#' @export
+
```
ChloeYou, line 8: Maybe an example would be helpful?

R/epi_shift_internal.R
```diff
@@ -0,0 +1,137 @@
...
+        default = default,
+        keys = keys,
+        columns = columns,
+        skip = skip,
+        id = id
+      )
+    )
+  }
+
+step_epi_shift_new <-
```
ChloeYou, line 49: What does step_epi_shift_new do? Does it create a new shifted(lag or lead) column?

ChloeYou: the name step_epi_shift_new was a bit confusing to me

tests/testthat/test-epi_shift_internal.R
```diff
@@ -0,0 +1,65 @@
+library(dplyr)
+library(epiprocess)
+library(parsnip)
+library(workflows)
+
```
ChloeYou, line 5: Maybe I didn't understand the new function clearly, but isn't `epi_shift_internal` or `epi_shift_new` the new function that automatically creates lags or leads based on the role of the variable? The idea is to unify the functions `step_epi_ahead` and `step_epi_lag` and test that unified functions, right?

tests/testthat/test-step_naomit2.R
```diff
@@ -0,0 +1,30 @@
...
+  step_epi_lag(death_rate, lag = c(0,7,14))
+
+z1 <- step_naomit2(r)
+z2 <- r %>%
+  step_naomit(all_predictors()) %>%
+  step_naomit(all_outcomes(), skip = TRUE)
+
+# Test
+test_that("Check that both functions behave the same way", {
+  expect_identical(z1$steps[[3]][-1][-5],z2$steps[[3]][-1][-5])
```
ChloeYou, line 26: Just out of curiosity, would it make sense to just call `expect_equal(z1, z2)` so that it covers the tests of the class type and the values in the tibble.

ChloeYou (commented):
Added some comments and questions about the code.

tests/testthat/test-step_naomit2.R
kenmawer: I tried that, but it doesn't work. The reason being is that each z1 and z2 contain some of their own unique identifiers, namely the quosure and the $id.

tests/testthat/test-epi_shift_internal.R
kenmawer: Yes, and we had step_epi_lag_new, step_epi_ahead_new, etc. which I also merged.

R/epi_shift_internal.R
dajmcdon: I agree that this nesting is a bit bizarre, but it's seemingly the [standard for step creation](https://www.tidymodels.org/learn/develop/recipes/). We'll be making a lot of these over the coming weeks, so it's probably a good time to read over these instructions.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,137 @@
...
+    columns = recipes_eval_select(x$terms, training, info),
+    skip = x$skip,
+    id = x$id
+  )
+}
+
+#' @export
+bake.step_epi_shift <- function(object, new_data, ...) {
+  if (!all(object$shift == as.integer(object$shift))) {
+    rlang::abort("step_epi_shift requires 'shift' argument to be integer valued.")
```
dajmcdon, line 86: The user isn't calling `step_epi_shift` so this error would be uninterpretable. It needs to report that the `ahead/lag` value should be integer valued. It's ok to have the same message for both uses, though, like "step_epi_ahead and step_epi_lag require `ahead`/`lag ..."

R/step_naomit2.R
dajmcdon: Yeah, this really needs a better name. I'm having less and less fondness for prefixing everything with `epi_`, but for now, maybe `step_epi_naomit()`

R/step_naomit2.R
```diff
@@ -0,0 +1,13 @@
+#' Special NA omit step that does two steps in one
+#'
+#' @param x Recipe to be used for omission steps
+#'
+#' @return Omits NA's from both predictors and outcomes and training time;
```
dajmcdon, line 5: "at training time"

tests/testthat/test-assign_arg_list.R
```diff
@@ -0,0 +1,6 @@
+test_that("First argument must be a list",{
+  expect_error(assign_arg_list(c(1,2,3)))
+})
+test_that("All arguments should be named",{
+  expect_error(assign_arg_list(list(1,2)))
+})
```
dajmcdon, line 6: Should also check if it "works" not just if it has errors. In other words, "do the assignments take place in the proper environment"?

tests/testthat/test-df_mat_mul.R
```diff
@@ -0,0 +1,19 @@
+df <- data.frame(matrix(1:100, ncol = 5))
+mat <- matrix(1:4, ncol = 2)
+df_mat_mul(df, mat, "z", dplyr::num_range("X", 2:3))
```
dajmcdon, line 3: What is this line doing outside of the test? (The first two create objects available to all the tests)

tests/testthat/test-df_mat_mul.R
```diff
@@ -0,0 +1,19 @@
...
+  expect_error(df_mat_mul(df, mat, "z", dplyr::num_range("X", 1:3)))
+})
+
+test_that("Matrix multiplication is being handled as expected", {
+  X <- df[c(1,4,5)]
+  Z <- as.data.frame(as.matrix(df[2:3]) %*% mat)
+  colnames(Z) <- c("z1","z2")
+  output <- cbind(X,Z)
+  expect_identical(df_mat_mul(df,mat, "z", dplyr::num_range("X", 2:3)),output)
+})
```
dajmcdon, line 19: * Maybe also a test to ensure that if we request columns without available names?
* Test to make sure that other `tidyselect` functions work.

tests/testthat/test-grab_names.R
```diff
@@ -0,0 +1,9 @@
+print("a")
```
dajmcdon, line 1: what's this for?

tests/testthat/test-step_naomit2.R
dajmcdon: We should at least add some comments here to explain why you need to grab these elements.

tests/testthat/test-step_naomit2.R
```diff
@@ -0,0 +1,30 @@
...
+  expect_identical(z1$steps[[4]][-1][-5],z2$steps[[4]][-1][-5])
+  expect_identical(class(z1$steps[[3]]),class(z2$steps[[3]]))
+  expect_identical(class(z1$steps[[4]]),class(z2$steps[[4]]))
```
dajmcdon, line 29: Again, why `[[4]]`?

See the comment above, but tests should ensure that other arguments don't break the functionality. What if someone calls `step_naomit2(..., someargs)`? The tests should cover anything that can possibly have the wrong downstream effects.

R/step_naomit2.R
kenmawer: Fixed this. I just got the idea of the "2" from how there also exists `read_csv2(...)`.

R/epi_shift_internal.R
kenmawer: @ChloeYou Yes, I just refactored the lag and ahead from the previous names of `step_epi_ahead_new` and `step_epi_lag_new`.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,137 @@
...
+#'
+#' @param shift A vector of integers. Each specified column will be
+#'  shifted for each value in the vector.
+#' @template step-return
+#'
+#' @details The step assumes that the data are already _in the proper sequential
+#'  order_ for shifting.
+#'
+#' @family row operation steps
+#' @export
```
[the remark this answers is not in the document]
kenmawer: Fixed.

tests/testthat/test-grab_names.R
kenmawer: Fixed, I put this on initially as a placeholder to save it onto my computer.

tests/testthat/test-assign_arg_list.R
kenmawer: Fixed.

tests/testthat/test-df_mat_mul.R
kenmawer: Fixed.

tests/testthat/test-df_mat_mul.R
kenmawer: How would I go about and do this when the third argument takes in a default, but the fourth argument is a ...?
For examples, `df_mat_mul(x,y,1:2)` assumes 1:2 is the third argument and not the ....

dajmcdon: Sorry, I forgot how this function works. So there's only one task really. What happens if you do:
```r
df_mat_mul(df, mat, "z", dplyr::num_range("Q", 2:3))
```

tests/testthat/test-step_naomit2.R
kenmawer: @dajmcdon Could you elaborate please? I don't see any tests that test for using the wrong number of arguments. However, I did add a check so that you must use a recipe for this to work.

I could also add a functionality that makes it permissible to use a data frame, whereby the data frame would be converted to a recipe.

dajmcdon: Oh, I see. `step_naomit()` accepts many args whereas, as written, `step_naomit2()` accepts only a single argument, the recipe. I don't think it needs more functionality. And actually, we don't need these tests (because it's just a wrapper around already tested functions). 

However, it would be nice to make that much more explicit in the documentation. Also, to be consistent, the argument should be `recipe` rather than `x`.

tests/testthat/test-step_naomit2.R
kenmawer: I just did.

tests/testthat/test-df_mat_mul.R
dajmcdon: Right, so there should probably be a test that expects an error.

tests/testthat/test-step_naomit2.R
kenmawer: I kept the tests anyway. However, I updated the documentation to be more specific and checked if the input is a recipe.

R/epi_lag.R
```diff
@@ -29,95 +29,17 @@ step_epi_lag <-
...
-        trained = trained,
-        lag = lag,
-        prefix = prefix,
-        default = default,
-        keys = keys,
-        columns = columns,
-        skip = skip,
-        id = id
-      )
+    step_epi_shift(recipe,
```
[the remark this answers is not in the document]
kenmawer: I did that.

R/epi_shift_internal.R
kenmawer: Fixed it to be comprehensible.

tests/testthat/test-epi_shift_internal.R
```diff
@@ -0,0 +1,65 @@
...
+              case_rate = rpois(100,20) + 1:200,
+              death_rate = rpois(100,10) + 1:200) %>%
+  as_epi_df()
+
+slm_fit <- function(recipe, data = x) {
+  workflow() %>%
+    add_recipe(recipe) %>%
+    add_model(linear_reg()) %>%
+    fit(data = data)
+}
```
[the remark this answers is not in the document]
kenmawer: I added tests to check for how the function works (ex: right number of predictors when doing stepping). I still don't quite get how I would check a function for fit aside from making sure it does the right number of steps and the linear function itself is right (rather than the outputs).

I also took out the randomness.

dshemetov:
Looks like this PR's commits are already in `frosting`? Not sure when they got in there, but we probably can just close this then.

Issue #45, opened 2022-06-10
get_test_data() should return an ungrouped epi_df


dajmcdon:
1. Ensure that the result of `get_test_data()` in ungrouped.
2. Add a test to check that this is true.

ChloeYou:
@dajmcdon do you mean to ungroup by geo_value?

ChloeYou:
I'll fix this right away! What kind of downstream issues would it cause, just curious?

dajmcdon:
Yes, exactly. Currently, the object that comes out is a grouped `epi_df` and it needs to be ungrouped.

(Also, it should really be grouped/sliced by geo_value and any `additional_metadata$other_keys`. You can use `epi_keys()` to grab the keys.)

Downstream, the grouping can break the more complicated behaviours needed at prediction time when you start doing the postprocessing. The grouping gums it up (hard for me to explain).

ChloeYou:
I see. Thanks for explaining! I'll change the grouping as well then.

ChloeYou:
@dajmcdon I don't think time_value should be included in the grouping right? or else the slice_tail is also grouped by time value.

PR #48, opened 2022-06-13
Fix grouping issue in test data

ChloeYou:
Closes https://github.com/cmu-delphi/epipredict/issues/45. 

Should this also be merged into the `dev` branch?


dshemetov:
I just switched this to the `frosting` branch. If we merge #49 in first (where `main` is merged into `frosting`), then the long commit log should disappear here.

brookslogan:
Right, this PR probably shouldn't be the one to merge `main` into `frosting`.  Taking a look at #49 to see if it can be merged in first.

brookslogan:
Also, I assume the merge conflicts GitHub is flagging now might automatically be resolved if #49 is merged.

R/get_test_data.R
```diff
@@ -42,14 +42,16 @@ get_test_data <- function(recipe, x){
...
+
   test_data <- x %>%
     dplyr::filter(
       dplyr::if_any(
         .cols = recipe$term_info$variable[which(recipe$var_info$role == 'raw')],
         .fns = ~ !is.na(.x)
       )
     ) %>%
-    epiprocess::group_by(geo_value) %>%
+    epiprocess::group_by(across(groups)) %>%
```
brookslogan, line 54: There's a tiny chance that the user will have defined a `groups` column, which will mess this up:
```
edf_with_groups_col = epiprocess::as_epi_df(tibble::tibble(geo_value="tx", groups=c("g1","g2","g3","g3","g3"), time_value=as.Date("2000-01-01")+0:4, value=1:5))
> edf_with_categories_col = epiprocess::as_epi_df(tibble::tibble(geo_value="tx", categories=c("g1","g2","g3","g3","g3"), time_value=as.Date("2000-01-01")+0:4, value=1:5))
> get_test_data(epi_recipe(edf_with_groups_col) %>% step_epi_ahead(value, ahead=0L) %>% step_epi_lag(value, lag=1L),
                edf_with_groups_col)
+ An `epi_df` object, with metadata:
* geo_type  = state
* time_type = day
* as_of     = 2022-06-15 06:18:59

# A tibble: 4 × 4
  geo_value time_value groups value
* <chr>     <date>     <chr>  <int>
1 tx        2000-01-01 g1         1
2 tx        2000-01-02 g2         2
3 tx        2000-01-04 g3         4
4 tx        2000-01-05 g3         5
> get_test_data(epi_recipe(edf_with_categories_col) %>% step_epi_ahead(value, ahead=0L) %>% step_epi_lag(value, lag=1L),
                edf_with_categories_col)
+ An `epi_df` object, with metadata:
* geo_type  = state
* time_type = day
* as_of     = 2022-06-15 06:19:25

# A tibble: 2 × 4
  geo_value time_value categories value
* <chr>     <date>     <chr>      <int>
1 tx        2000-01-04 g3             4
2 tx        2000-01-05 g3             5
```

It looks like using `across(tidyselect::all_of(groups))` would prevent this.

ChloeYou: Good catch and thank you for illustrating it with the example! I've updated the line now.

R/get_test_data.R
```diff
@@ -42,14 +42,16 @@ get_test_data <- function(recipe, x){
...
+
   test_data <- x %>%
     dplyr::filter(
       dplyr::if_any(
         .cols = recipe$term_info$variable[which(recipe$var_info$role == 'raw')],
         .fns = ~ !is.na(.x)
       )
     ) %>%
-    epiprocess::group_by(geo_value) %>%
+    epiprocess::group_by(across(dplyr::all_of(groups))) %>%
```
brookslogan, line 54: Looks like `across` hasn't been `@importFrom`'d, so we need `dplyr::` as well.  If you run `devtools::check()` or `devtools::check(vignettes=FALSE)` you will see
```
   get_test_data: no visible global function definition for ‘across’
```

CI should be automatically flagging this + a lot of pre-existing issues; maybe we haven't triggered it yet since it'd been set up, or maybe there's a configuration issue or something deliberately disabled.

brookslogan: The reason why `check` is complaining is because this looks like a situation where
- if the user doesn't have a package attached with `across`, this would be an error
- if the user has a package attached with a different `across` function than we wanted, or defined their own `across` function as a global, this might have different behavior than wanted
This is because when a package doesn't import or use `::` on all its dependencies, R will fall back on whatever the user has defined globally or is available in attached packages.

Why does the test still pass, even if we try defining `across=function(...) stop("using a different across than intended")`?  

Short answer: because `dplyr` saves us.

Long answer: there are a few things that each would prevent some/all issues:
- `dplyr` saves us: it appears to add `across` as part of `.data` (within `group_by`), which is searched before `.env` which would contain the custom `across` (we'll find an `across` function, and it's `dplyr`'s)
- the test file `library`s in (loads and attaches) `dplyr` (we'll find an `across`, the right one unless we define our own `across` or attach another package with a different `across` --- e.g., we would get "using a different across than intended")
- the `recipes` dependency has `Depends: dplyr` for some reason, which attaches `dplyr` (we'll find an `across`, but not necessarily the right one --- this alone wouldn't prevent the custom-`ahead` issue --- e.g., we would get "using a different across than intended")

We still want to add the scope or `@importFrom` to:
- satisfy the package checks
- hedge against `dplyr` removing `across` from `.data`

brookslogan: Editing the above to fix some notes.

brookslogan (commented):
Looks good to resolve #45.  I have some other questions and comments about how the function is supposed to work, but I will add them to a separate issue.

brookslogan (approved):

Issue #36, opened 2022-06-05
step_training_window


dajmcdon:
Create a preprocessing step that limits the size of the training window to the `n` most recent observations per location.

PR #53, opened 2022-06-17
Created a preprocessing step that limits the size of the training window

rachlobay:
Addressed [issue 36](https://github.com/cmu-delphi/epipredict/issues/36) by creating a preprocessing step that limits the size of the training window to the n most recent observations per location. See training_window.R and test-step_training_window.R for some tests.


R/training_window.R
```diff
@@ -0,0 +1,109 @@
...
+#'
+#' @param recipe A recipe object. The step will be added to the
+#'  sequence of operations for this recipe.
+#' @param ... One or more selector functions to choose variables
+#'  for this step. See [selections()] for more details.
+#' @param role For model terms created by this step, what analysis role should
+#'  they be assigned?
+#' @param trained A logical to indicate if the quantities for
+#'  preprocessing have been estimated.
+#' @param nrec An integer value that represents the number of most recent
```
dajmcdon, line 15: Maybe rename this to `n_recent`, just for clarity.

R/training_window.R
```diff
@@ -0,0 +1,109 @@
+#' Limits the size of the training window to the most recent observations
+#'
+#' `step_training_window` creates a *specification* of a recipe step that
+#'   limit the size of the training window to the `nrec` most recent
+#'   observations in `time_value` per location from `geo_value`.
+#'
+#' @param recipe A recipe object. The step will be added to the
+#'  sequence of operations for this recipe.
+#' @param ... One or more selector functions to choose variables
```
dajmcdon, line 9: Do we actually need this argument? I don't think we need `role` either.

Basically, this operates on the `time_value` (and any keys) column right? It doesn't need anything else?

dajmcdon: It's probably good to make this resemble [`step_slice()`](https://recipes.tidymodels.org/reference/step_slice.html), since that's basically what we're doing.

dajmcdon: So, I'm thinking just say that `role` and `...` are unused.

R/training_window.R
```diff
@@ -0,0 +1,109 @@
+#' Limits the size of the training window to the most recent observations
+#'
+#' `step_training_window` creates a *specification* of a recipe step that
+#'   limit the size of the training window to the `nrec` most recent
+#'   observations in `time_value` per location from `geo_value`.
+#'
```
dajmcdon, line 6: We probably need some sort of warning to indicate that `step_epi_lead()` and `step_epi_lag()` should come AFTER any filtering.

R/training_window.R
```diff
@@ -0,0 +1,109 @@
...
+           skip = TRUE,
+           id = rand_id("training_window")) {
+
+    add_step(
+      recipe,
+      step_training_window_new(
+        role = role,
+        trained = trained,
+        nrec = nrec,
+        skip = skip,
```
dajmcdon, line 58: We should just set `skip = TRUE` here to make sure it never occurs on the test data.

R/training_window.R
```diff
@@ -0,0 +1,109 @@
...
+#' @param ... One or more selector functions to choose variables
+#'  for this step. See [selections()] for more details.
+#' @param role For model terms created by this step, what analysis role should
+#'  they be assigned?
+#' @param trained A logical to indicate if the quantities for
+#'  preprocessing have been estimated.
+#' @param nrec An integer value that represents the number of most recent
+#' observations that are to be kept in the training window per location.
+#' The default value is 50.
+#' @param skip A logical. Should the step be skipped when the
```
dajmcdon, line 18: The step should always be skipped on the test data.

R/training_window.R
```diff
@@ -0,0 +1,109 @@
...
+  )
+}
+
+#' @export
+bake.step_training_window <- function(object, new_data, ...) {
+  if (!all(object$nrec == as.integer(object$nrec))) {
+    rlang::abort("step_training_window requires 'nrec' to be integer valued.")
+  }
+
+  new_data %>% dplyr::group_by(geo_value) %>%
```
dajmcdon, line 94: This likely needs to handle any additional keys as well. Calling `epi_keys()` should give you all the columns.

dajmcdon (changes requested):

R/training_window.R
rachlobay: Ok. Changed all instances of `nrec` --> `nrecent`

R/training_window.R
rachlobay: Ok. I removed `...` with no complaints from RStudio, But when I remove `role`, I get an error after using prep that says "Error in if (!is.na(x$steps[[i]]$role)) { : argument is of length zero". Based on that, I think we should keep role even if it isn't used. Also, it is kept in `step_slice()`, so I think it is probably ok to also keep here. I'll just note in the documentation that it is unused, as you said.

R/training_window.R
rachlobay: Done.

R/training_window.R
rachlobay: Right... So I've just removed `skip` from the user-facing parameters and that documentation for it.

R/training_window.R
rachlobay: Added a note for this in `@details` part of the documentation.

R/training_window.R
rachlobay: And to address your 2nd q. from your first comment on this code, the fun. needs `time_value` and `geo_value` columns as the issue said to limit the training window n most recent observations per location.

R/training_window.R
rachlobay: @dajmcdon Could you elaborate on the above comment? What do you mean by handle any additional keys (for ex., do you mean to allow for them in `group_by`?

rachlobay: Updated w/slight modification to ensure that `group_by` works as intended (that is, `group_by(across(all_of(ek)))`)

dajmcdon:
It looks like this is failing because of cmu-delphi/epiprocess#96. Blocked for the moment.

rachlobay:
@dajmcdon, when using training data in `bake()` (that is, `bake(new_data = NULL)` such as in the tests for this function in particular, we now lose the `epi_df` class of the resulting data frame. In that case, the baked data gets demoted to a plain old tibble. 

I think the issue here is that when we put `new_data = NULL` in bake, `bake.recipe` sends us to `juice()` and since the composition is a tibble (by default), then `new_data <- tibble::as_tibble(new_data)` on line 829 of `juice()` here: https://github.com/tidymodels/recipes/blob/main/R/recipe.R. This is why I think that all was well with the class when I used `bake(new_data = NULL)` along w/that old `as_tibble.epi_df` function that was removed… What do you think?

dajmcdon:
I suspect you're right. We're going to need a new `bake.epi_recipe()` method. Ugh.

Issue #38, opened 2022-06-05
post_add_forecast_date(), *_add_target_date()


dajmcdon:
Create a postprocessing step that adds the forecast date and the target date.

brookslogan:
Sorry for the confusion when discussing this at the meeting.  From some more inspection of the `simple-forecasts.Rmd`:
- Yes, with `as_of=TRUE`, the forecast date is `time_value`, and the target date is `time_value+ahead`.  (Although, personally, I think it's a bit confusing that `epix_slide` calls the column that should be the forecast date the `time_value` rather than the `version`.  In fact, when we think about composability, I think it will be more natural to call the target date the `time_value`, because then we can think of performing backcasting on data signals and calibration on forecasts in the same sort of way.  May file another issue.)
- With `as_of=FALSE` I think this is also the assumption we are making, although it doesn't really make sense with real data sources without augmentation with nowcasts.
But:
- Note #50; I believe the above settings are what things "should" be, but they will be wrong / `arx_forecaster` will be suboptimal until this bug is fixed.

PR #55, opened 2022-06-18
Created post processing steps for adding the forecast and target dates

rachlobay:
Addressed [Issue #38](https://github.com/cmu-delphi/epipredict/issues/38) by making `layer_add_forecast_date()` and `layer_add_target_date()` as well as tests for each. I wasn't 100% sure about what was intended for the target date, so I defined it as in the simple-forecasts.Rmd vignette (where `target_date = time_value + ahead`), but perhaps that wasn't quite what was intended... Let me know and I can update it.

**Update:** Currently, I don't think the preprocessor is passed along (so that, for ex., `ahead` doesn't need to be provided by the user to be accessed in these postprocessing functions)... Please hold off on reviewing for now until that is done and I have updated these two functions accordingly.


rachlobay:
Currently, I don't think the preprocessor is passed along (so, for ex., `ahead`doesn't need to be provided by the user to be accessed)... Please hold off on reviewing until that is done.

R/layer_add_forecast_date.R
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+layer_add_forecast_date_new <- function(forecast_date, newdata, id = id) {
+  layer("add_forecast_date", forecast_date = forecast_date, newdata = newdata, id = id)
+}
+
+slather.layer_add_forecast_date <- function(object, components, ...) {
+  if (is.null(object$newdata)) stop("`newdata` must be specified as an argument.")
+
+  last_rows_df <- object$newdata %>% # %% new_data for forecast_date needed here... perhaps this is components or components$forged$predictors here???
```
dajmcdon, line 79: Yeah, the `newdata` will be available in `components`. 

I think you also need to `@export` this function.

R/layer_add_forecast_date.R
```diff
@@ -0,0 +1,106 @@
...
+    as.Date(stringr::str_extract(
+      attributes(object$newdata)$metadata$as_of, # %% new_data change
+      "\\d{4}-\\d{2}-\\d{2}"
+    ))
+
+  ahead <- as.numeric(stringr::str_extract(names(components$mold$outcomes),
+                                           "(?<=ahead_)\\d+"))
+  if (is.null(object$forecast_date)) {
+  object$forecast_date <- max_time_value + ahead
+  warning("Set forecast_date equal to maximum time value plus ahead value.")
```
dajmcdon, line 95: let's indent these for clarity.

R/layer_add_target_date.R
```diff
@@ -0,0 +1,54 @@
...
+    )
+  }
+
+layer_add_target_date_new <- function(id = id) {
+  layer("add_target_date", id = id)
+}
+
+#' @export
+slather.layer_add_target_date <- function(object, components, ...) {
+  ahead <- as.numeric(stringr::str_extract(names(components$mold$outcomes),
```
dajmcdon, line 47: Here, I think you can read off the recipe.

dajmcdon (changes requested):
I think this review is perhaps still premature? Ignore if so.

R/layer_add_forecast_date.R
```diff
@@ -0,0 +1,106 @@
...
+}
+
+slather.layer_add_forecast_date <- function(object, components, ...) {
+  if (is.null(object$newdata)) stop("`newdata` must be specified as an argument.")
+
+  last_rows_df <- object$newdata %>% # %% new_data for forecast_date needed here... perhaps this is components or components$forged$predictors here???
+    dplyr::group_by(geo_value) %>%
+    dplyr::slice(dplyr::n())
+
+  max_time_value <- max(last_rows_df$time_value)
```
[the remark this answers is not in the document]
rachlobay: You're right. I agree that `max()` is enough (the other part is unnecessary).

R/layer_add_forecast_date.R
```diff
@@ -0,0 +1,106 @@
...
+
+  as_of_date <-
```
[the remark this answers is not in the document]
rachlobay: Ok. I think I see... Used `as.Date.POSIXct()` to extract YYYY-MM-DD from `as_of`

R/layer_add_forecast_date.R
```diff
@@ -0,0 +1,95 @@
...
+
+  max_time_value <- max(components$keys$time_value)
+
+  as_of_date <- as.Date.POSIXct(attributes(components$keys)$metadata$as_of)
+
+  ahead <- the_recipe$steps[[2]][["ahead"]]
+
+  if (is.null(object$forecast_date)) {
+    object$forecast_date <- max_time_value + ahead
+    warning("Set forecast_date equal to maximum time value plus ahead value.")
```
dajmcdon, line 84: Let's not produce this warning, but emphasize it in the documentation (you already say it in Details, but let's add it to the @param part on line 4.

R/layer_add_forecast_date.R
```diff
@@ -0,0 +1,95 @@
...
+layer_add_forecast_date_new <- function(forecast_date, id = id) {
+  layer("add_forecast_date", forecast_date = forecast_date, id = id)
+}
+
+#' @export
+slather.layer_add_forecast_date <- function(object, components, the_fit, the_recipe, ...) {
+
+  max_time_value <- max(components$keys$time_value)
+
+  as_of_date <- as.Date.POSIXct(attributes(components$keys)$metadata$as_of)
```
dajmcdon, line 78: just `as.Date()` is enough. The rest is the S3 method since `class(attributes(components$keys)$metadata$as_of) == "POSIXct"`.

DESCRIPTION
```diff
@@ -25,6 +25,7 @@ Imports:
     purrr,
     recipes (>= 0.2.0.9001),
     rlang,
+    stringr,
```
dajmcdon, line 28: Let's remove this since we're not using it anymore.

R/layer_add_target_date.R
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+layer_add_target_date_new <- function(id = id) {
+  layer("add_target_date", id = id)
+}
+
+#' @export
+slather.layer_add_target_date <- function(object, components, the_fit, the_recipe, ...) {
+  the_fit <<- the_fit
+  the_recipe <<- the_recipe
```
dajmcdon, line 48: this part was just for testing right?

R/layer_add_target_date.R
```diff
@@ -0,0 +1,55 @@
...
+#' f <- frosting() %>% layer_predict() %>%
+#'   layer_add_target_date() %>% layer_naomit(.pred)
+#' wf1 <- wf %>% add_frosting(f)
+#'
+#' p <- predict(wf1, latest)
+#' p
+layer_add_target_date <-
+  function(frosting, id = rand_id("add_target_date")) {
+    add_layer(
+      frosting,
```
dajmcdon, line 34: I think we should actually have a `target_date` argument. Let's have it be NULL by default. That way it's possible for the user to override the detection based on the `ahead` value in the recipe.

dajmcdon (changes requested):
You'll have to merge frosting to get the tests to pass.

R/layer_add_target_date.R
rachlobay: Whoops... Accidentally left that in there. It is now removed.

R/layer_add_target_date.R
rachlobay: Ok. I've done so... Do you want any warnings if someone specifies an unusual target date or nah?

dajmcdon: Let's say... "nah"

R/layer_add_forecast_date.R
```diff
@@ -2,18 +2,20 @@
 #'
 #' @param frosting a `frosting` postprocessor
 #' @param forecast_date The forecast date to add as a column to the `epi_df`.
```
dajmcdon, line 4: Oops. Sorry, I missed this. To my mind the `forecast_date` is the date on which the forecast is made. So, by default, it should be `max(time_value)` from the training data. The `target_date` should be "the date the forecast is for". So that one should be `max(time_value) + ahead` by default. 

It looks like they're both the same currently right?

dajmcdon (changes requested):

R/layer_add_forecast_date.R
rachlobay: Almost… Current defaults: `forecast date = max_time_value + ahead` (where `max_time_value` is for test data), `target_date = time_value + ahead` (based on simple.forecasts.Rmd). But I will change to what you specified. A couple qs for that...

For `forecast_date`, to get the max time value from the training data, is that using mold from components? Then, I check if the `forecast_date < as_of_date` of the test data and throw a warning there (that says "forecast_date is less than the most recent update date of the data.”, yes?

For `target_date`, by default, that is the max time value in the in the **test** data + ahead or no?

dajmcdon: The difficulty here is that leading/lagging increments the `time_value`. So this could all be a bit dangerous. In an ideal setting:
1. `target_date` would be the max time value in the test data + ahead, as you said.
2. `forecast_date` would be the max time value in the test data. (So seemingly, the date of the most recent data available to you). Acting like we have data up to (and including) today, and we produce a forecast today, then this should work.
However, lots of weird crud can happen that would screw these up. If data isn't available for today but only for yesterday, then that would throw things off. If we accidentally lead the test data, then it'll produce "future" time_values.

dajmcdon (approved):

PR #59, opened 2022-06-21
Residual quantiles added as postprocess layer

ChloeYou:
Somehow the layer_predict unit tests are giving 3 rows instead of 108 rows. Testing if it runs properly on a virtual machine.


tests/testthat/test-layer_residual_quantile.R
```diff
@@ -0,0 +1,27 @@
...
+  step_epi_lag(death_rate, lag = c(0, 7, 14)) %>%
+  step_epi_ahead(death_rate, ahead = 7) %>%
+  step_naomit(all_predictors()) %>%
+  step_naomit(all_outcomes(), skip = TRUE)
+
+wf <- epi_workflow(r, parsnip::linear_reg()) %>% fit(jhu)
+latest <- get_test_data(recipe = r, x = jhu)
+
+
+test_that("function works", {
```
dajmcdon, line 14: let's give this test a slightly more informative name.

dajmcdon (approved):
Everything else looks good.

I think the reason for 3 rows is because you used `layer_naomit()`. So this is actually as designed!

@ChloeYou After making the minor change, feel free to merge.

PR #61, opened 2022-06-21
Frosting cleanup

dajmcdon:
* Exports and documents the main frosting components.
* Tests that frosting can be added/removed.


dajmcdon:
@ChloeYou @rachlobay If one of you can approve and merge, then you don't need to prefix these in your other open PRs.

PR #67, opened 2022-06-24
64 check frosting has predict layer 

### LINKED ISSUE #64: Check `frosting` has layer of class `predict` in `apply_frosting.epi_workflow()`
Check if the prediction layer is missing and do it first, also consider adding some reasonable default.

The idea is that in `apply_frosting.epi_workflow()`, we:
1. Check the frosting to see if one of the layers is of class predict.
2. If 1 is TRUE do nothing.
3. If 1 is FALSE, do it first (maybe just add layer_predict() on to the top of the layer stack automatically.

ChloeYou:
Closes #64.

1. Adds layer_predict() to workflow if it is missing. It alters the workflow object permanently and adds the layer if the user forgot to add it. 
2. One unit test added.
3. There seems to be an error in the current frosting branch. Fixes have been made to allow for the checks to run. Worth noting that real df has geo_value reported as the first column. currently the package doesn't enforce time_value to be the first column. Some decisions to be made here I think.


R/frosting.R
```diff
@@ -194,6 +194,20 @@ apply_frosting.epi_workflow <- function(workflow, components, the_fit, ...) {
     return(components)
   }
   layers <- workflow$post$actions$frosting$frosting$layers
+
+  # checks if layer_predict() is in the postprocessors
+  list_of_layer_names <- c()
+  for (l in seq_along(layers)){
+    la <- layers[[l]]
+    list_of_layer_names <- append(list_of_layer_names,class(la)[1])
+  }
```
dajmcdon, line 203: While this "works", in R (and other languages), this type of vector-growing-in-a-loop should be avoided at all costs. The reason is that, internally, R has to sequentially hunt for memory at each iteration.

Best practice is:
1. Avoid the loop if possible.
2. If not possible, allocate space for the whole vector before the loop.

To do 2, one would just do `list_of_layer_names <- character(length(layers))`. But 1 is possible. Try replacing these lines with:
```r
layer_names <- map_chr(layers, ~ class(.x)[1])
```

For example:
``` r
A = matrix()
B = data.frame()
C = vector()
D = factor()
l = list(A=A, B=B, C=C, D=D)

map_chr(l, ~ class(.x)[1])
#>            A            B            C            D 
#>     "matrix" "data.frame"    "logical"     "factor"
```

<sup>Created on 2022-06-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

dajmcdon (changes requested):
Everything else looks good.

dajmcdon (approved):

PR #77, opened 2022-06-28
km-kill-lags branched off frosting.

kenmawer:
Closes issue #22.


dajmcdon:
@kenmawer I'm not sure what the error is here, but this needs some debugging.

kenmawer:
@dajmcdon These errors seem to be coming from tests of functions such as those relating to frosting, none of which I worked on.

dajmcdon:
Right, but those errors come from tests that use `step_epi_lag()`. So something that this branch does broke those tests. Two things I saw:

* Calling `get_test_data()` no longer works.
* In `test-frosting.R` on line 46, the predictions have the wrong date.

kenmawer:
Still needs work.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,225 @@
...
+step_epi_lag <-
+  function(recipe,
+           ...,
+           role = "predictor",
+           trained = FALSE,
+           lag = 1,
+           default = NA,
+           keys = epi_keys(recipe),
+           columns = NULL,
+           skip = FALSE) {
```
dajmcdon, line 62: Despite what I said some weeks ago, we should have `prefix` and `id` here. That's just to be consistent with the typical `recipes::step_*()` functions.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,225 @@
...
+    if (any(lag<0)) {
```
dajmcdon, line 63: Let's actually make this an error.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,225 @@
...
+step_epi_ahead <-
+  function(recipe,
+           ...,
+           role = "outcome",
+           trained = FALSE,
+           ahead = 1,
+           default = NA,
+           keys = epi_keys(recipe),
+           columns = NULL,
+           skip = FALSE) {
```
dajmcdon, line 95: Same comment here about the arguments.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,225 @@
...
+    if (any(ahead<0)) {
```
dajmcdon, line 96: and again, error here.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,225 @@
...
+    skip = x$skip,
+    id = x$id
+  )
+}
+
+#' @export
+bake.step_epi_shift <- function(object, new_data, ...) {
+  is_lag <- object$prefix == "lag_"
+  if (!all(object$shift == as.integer(object$shift))) {
+    error_msg <- paste0("step_epi_",
```
dajmcdon, line 181: Let's actually check the integer values on creation rather than at bake time.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,225 @@
...
+    dplyr::group_by(dplyr::across(dplyr::all_of(ok[-1]))) %>%
+    dplyr::arrange(time_value) %>%
+    dplyr::ungroup()
+
+}
+
+#' @export
+print.step_epi_shift <-
+  function(x, width = max(20, options()$width - 30), ...) {
+    ## TODO add printing of the shifts
```
dajmcdon, line 220: Is this TODO completed?

dajmcdon (changes requested):

R/epi_shift_internal.R
kenmawer: I get why this would be the case. However, there would be no infallible way of checking whether something is a lead or lag based on print statements, as a 0 can be a lag or lead here. Should we have that the 0 be treated as a lag, as it is a predictor (given that I am to have errors thrown for negative values)?

dajmcdon: If the user calls `step_epi_lag()`, then the behaviour is to lag. So the prefix should be `lag_`. Same for `step_epi_lead()`. Then that gets passed directly to `step_epi_shift()`.

kenmawer: That is only the default behaviour of the function. However, one could put in `prefix = "asdf_"` or something. If there is no `prefix` argument for lag, then the prefix will be forced as `lag_`.

dajmcdon: Sorry, maybe there's some misunderstanding. Please put those arguments (with the defaults) into the `step_epi_lag()` and `step_epi_ahead()` functions rather than omitting them there but calling `step_epi_shift()` with those arguments.

R/epi_shift_internal.R
kenmawer: Yes.

R/epi_shift_internal.R
```diff
@@ -0,0 +1,218 @@
...
+           ahead = 1,
+           prefix = "ahead_",
+           default = NA,
+           keys = epi_keys(recipe),
+           columns = NULL,
+           skip = FALSE,
+           id = rand_id("epi_ahead")) {
+
+    stopifnot("Ahead values must be positive integers" =
+                all(ahead>0 & ahead == as.integer(ahead)))
```
dajmcdon, line 100: `ahead == 0` should be ok.

dajmcdon (changes requested):
Note that you can safely edit this branch locally and push to the same remote.

R/epi_shift_internal.R
kenmawer: I intentionally did `ahead > 0` as that'd be the only way to unambiguously distinguish a `lag` from an `ahead` given that it's possible to change the names of the parameters, but that the shift value distinguishes between the two.

How should I go about it? I tried doing many different things, but this should be fail-safe while satisfying guidelines.

dajmcdon:
If you merge `frosting` onto this branch, then the checks should pass.

dajmcdon:
@kenmawer I fixed up the printing to work the way we discussed in my office. Why don't you pull locally, run the examples, and see what you think. Have a look at how I did it just to make sure that it makes sense. Once it does, you can approve this PR and merge.

dajmcdon (approved):

PR #79, opened 2022-06-28
step_epi_naomit branched off frosting.

kenmawer:
This is taken from a messy pull request, as if it were broken up.


R/step_epi_naomit.R
```diff
@@ -0,0 +1,20 @@
...
+#'   outcomes at prediction time due to lack of response and avoidance
+#'   of data loss.
+#' @export
+#' @examples
+#' case_death_rate_subset %>%
+#'  epi_recipe() %>%
+#'  step_epi_naomit()
+
+step_epi_naomit <- function(recipe) {
+  stopifnot("recipe" %in% class(recipe))
```
dajmcdon, line 16: better is
```r
stopifnot(inherits(x, "recipe")
```

dajmcdon (changes requested):
I'm not sure why the Rd files for create_layer and new_epi_recipe_blueprint are here. (They shouldn't be). Maybe rerun `devtools::document()`?

R/step_epi_naomit.R
kenmawer: I rerun `devtools::document` and the Rd files for create_layer and new_epi_recipe_blueprint haven't been deleted.
Also, I updated the second comment.

dajmcdon (approved):

Issue #63, opened 2022-06-22
Distributional forecasts


dajmcdon:
Once #59 is merged, we'll have the first example of distributional predictions. These can also be created using the `pred_int` argument of `layer_predict()` do get parametric intervals. 

We should examine using the [distributional](https://pkg.mitchelloharawild.com/distributional/) package to unify behaviour across different types of forecasts. This would allow for a 1-column specification rather than column-per-quantile as currently. It can also handle, say, Gaussian intervals via mean+variance. This would allow for extraction down the line.

Issue #46, opened 2022-06-11
Allow `layer_predict()` to pass along `...`


dajmcdon:
On the `frosting` branch, there's a `layer_predict()` function. The dots need to be converted to a list, then passed back to the `predict()` call in the `slather` method. I tried a few different implementations but couldn't get it to work. 

* In the user facing `layer_predict()` function, the dots must be converted as far as I can tell (not simply passed along). This is because each layer is saved into a list. The current implementation is fine here.
* But the `dots_list` would then need to be evaluated/defused on line 37. I tried combinations of `rlang` things. I also tried using `do.call()` with a list of args. I couldn't make either work (and the do.call version would seem to have undesirable copy behaviour).

@brookslogan I'm assigning you since you seem to have the most facility with quosure behaviour. I suspect there's a simple fix, but it's beyond me.

brookslogan:
Sorry, just getting to this.
- Can the dots be immediately evaluated in `layer_predict`?  If not, can they be evaluated multiple times?
- Do you want the dots to be evaluated like normal arguments, or do you need to manipulate the environments or make `.data` and `.env` pronouns available to the user?

A few guesses at what would work:
- Immediately evaluate like normal arguments: use `list`, `list2`, or `dots_list` in `layer_predict`.  In `slather` use `rlang::inject(stats::predict(<stuff>, !!!object[[["dots_list"]]))`.
- Evaluate on slather, potentially multiple times, like normal arguments: use `enquos0` or `enquos`.  In `slather` use `rlang::inject(stats::predict(<stuff>, !!!lapply(object[["dots"]], rlang::eval_tidy)))`.
- Evaluate on slather, potentially multiple times, with environment manipulation or pronouns: same as previous bullet point but with appropriate arguments to `eval_tidy`.
- Evaluate on slather, don't evaluate multiple times, with/without environment manipulation: we'd need a (custom?) approach where the `dots` would be an environment with some promise (`delayedAssign`) bindings.

dajmcdon:
Your comment illustrates exactly how little I know about this. The dots should be "additional named arguments passed on to the predict method". So I think the answers are:

* Yes? Not evaluated multiple times.
* Normal arguments?

I suspect this means that Bullet 1 should work (or something like it). However I tried some simpler versions, and ... no dice.

[...]

PR #88, opened 2022-07-06
Add distn

dajmcdon:
Partially addresses #63 .

1. Requires adding some post-processing layers to add the quantiles.
2. Suggest we remove the `...` args in `layer_predict()` (see #46 ) and prefer explicit quantile addition. This would use the `slather` method to process the output as needed (perhaps depending on the fitted model).


R/layer_residual_quantiles.R
```diff
@@ -0,0 +1,88 @@
+#' Creates predictions based on residual quantiles
+#'
+#' @param frosting a `frosting` postprocessor
+#' @param ... Unused, include for consistency with other layers.
+#' @param probs numeric vector of probabilities with values in (0,1)
+#'   referring to the desired quantile.
+#' @param symmetrize logical. If `TRUE` then interval will be symmetric.
+#' @param name character. The name for the output column.
+#' @param .flag a logical to determine if the layer is added. Passed on to
```
ChloeYou, line 9: Is `.flag` still preserved or did we agree to remove it?

dajmcdon: It's there for now, just for consistency with the other `layer_*()` functions. But the plan is to change the name later. (Separate PR).

R/dist_quantiles.R
```diff
@@ -0,0 +1,265 @@
...
+  tau <- as_list_of(tau, .ptype = double())
+  args <- vec_recycle_common(x = x, tau = tau)
+  qntls <- as_list_of(map2(args$x, args$tau, new_quantiles))
+  new_vctr(qntls, class = "distribution")
+}
+
+
+# format.dist_quantiles <- function(x, ...) {
+#   sprintf("quantiles[%s]", length(x))
+# }
```
mgyliu, line 65: Was this commented out section left here on purpose? If so, maybe good to document what it's for

dajmcdon: It was an earlier version that doesn't work correctly. Probably should remove.

PR #93, opened 2022-07-08
Km test data

kenmawer:
Added a test to `data.R` to make sure the dataset is what it claims to be.


dajmcdon:
Thanks for checking, but the package doesn't actually need such a test. I'm going to close without merging.

PR #96, opened 2022-07-14
Grab layer args

dajmcdon:
You should be able to use this function to "grab" the arguments passed to a `layer_*()` or `step_*()` function.

Important notes:
1. It doesn't really grab the arguments, but rather the stored objects. So `step_epi_ahead(..., ahead = 7)` actually stores `shift = -7`. If you ask for the argument named `ahead`, you'll get an error. See the tests.
2. If your `layer`/`step` name matches multiple layers or steps, you'll get a list. Otherwise, you'll get the object.


dajmcdon:
@rachlobay once #97 is merged, you should be able to get the the right stuff from `step_epi_lag()`/`step_epi_ahead()` again.

PR #97, opened 2022-07-16
finish revert

dajmcdon:
~~Two tests fail on check when run locally with `devtools::check():~~
* ~~test-epi_keys~~
* ~~test-blueprint~~

~~But if I just run those tests individually, nothing fails. I can't figure out the problem.~~

@kenmawer have a look, and comment+approve+merge when you're satisfied.


kenmawer:
The naming is inconsistent: the name of the `epi_shift_internal` function has been changed to `step_epi_shift`; however, the name of the test is `test-epi_shift_internal`.
Where does the `print_step_shift` come from? It's not a defined method in `step_epi_shift`.

R/step_epi_shift.R
```diff
@@ -229,22 +216,46 @@ bake.step_epi_shift <- function(object, new_data, ...) {
...
-print.step_epi_shift <- function(x,
-                                 width = max(20, options()$width - 30),
-                                 title = "Shifting ",
-                                 shift = x$shift,
-                                 ...) {
-  print_step_shift(x$columns, x$terms, x$trained, title, width, shift = shift)
+print.step_epi_ahead <- function(x, width = max(20, options()$width - 30), ...) {
+  print_step_shift(x$columns, x$terms, x$trained, "Leading ", shift = x$ahead)
   invisible(x)
 }
```
kenmawer, line 260: Where does `print_step_shift` come from?

kenmawer (changes requested):

R/step_epi_shift.R
dajmcdon: If you click the little blue arrow just 👇 , it should expand the script.

If the code is already on the target branch, it won't be shown in the PR. This function isn't changed by the PR.

dajmcdon:
By "naming inconsistent" do you mean that the `*_ahead()`/`*_lag()` should be in different files (with matching names)? And rename the `test-*` file?

kenmawer:
> By "naming inconsistent" do you mean that the `*_ahead()`/`*_lag()` should be in different files (with matching names)? And rename the `test-*` file?

No, I don't mean that.

My point is that the name for the file is `step_epi_shift`, but that the name for the corresponding test is `test-epi_shift_eternal` rather than `test-step_epi_shift`.

dajmcdon:
Ok, done!

kenmawer (approved):
Looks good to me!

EDIT: Hold on--Do not merge now, I didn't notice something until now.

R/step_epi_shift.R
```diff
@@ -180,30 +148,49 @@ step_epi_shift_new <-
...
+
+
 #' @export
-prep.step_epi_shift <- function(x, training, info = NULL, ...) {
-  step_epi_shift_new(
+prep.step_epi_lag <- function(x, training, info = NULL, ...) {
+  step_epi_lag_new(
     terms = x$terms,
     role = x$role,
     trained = TRUE,
```
kenmawer, line 158: Should be `trained = x$trained`.

R/step_epi_shift.R
```diff
@@ -180,30 +148,49 @@ step_epi_shift_new <-
...
-bake.step_epi_shift <- function(object, new_data, ...) {
-  grid <- tidyr::expand_grid(col = object$columns, shift_val = object$shift) %>%
-    dplyr::mutate(newname = glue::glue(
-                          paste0("{object$prefix}","{abs(shift_val)}","_{col}")
-                        )
+prep.step_epi_ahead <- function(x, training, info = NULL, ...) {
+  step_epi_ahead_new(
+    terms = x$terms,
+    role = x$role,
+    trained = TRUE,
```
kenmawer, line 174: Should be `trained = x$trained`.

kenmawer (changes requested):
a

R/step_epi_shift.R
dajmcdon: I actually think this is correct based on [recipes::prep.step_lag()](https://github.com/tidymodels/recipes/blob/93910b03a67a6243e52342d9b246ce30e72a0a8f/R/lag.R#L88).

kenmawer: I stand corrected then.

kenmawer (approved):
Looks good now. Things might need to be changed should this be updated, such as the part with `case_weights`.

PR #100, opened 2022-07-18
40 steppost population scaling

ChloeYou:
- [x] Preprocessing function completed and unit tests added
- [x] Postprocessing function and unit tests


R/layer_population_scaling.R
```diff
@@ -0,0 +1,133 @@
...
+#' @param frosting a `frosting` postprocessor
+#' @param ... the column(s) in the `epi_df` to scale back.
+#' @param df a data frame that contains the population data used for scaling.
+#' @param by A character vector of variables to join by. Column names should be
+#' in lower case.
+#' To join by different variables on the `epi_df` and `df`, use a named vector.
+#' For example, by = c("geo_value" = "states") will match `epi_df$geo_value`
+#' to `df$states`. To join by multiple variables, use a vector with length > 1.
+#' For example, by = c("geo_value" = "states", "county" = "county") will match
+#' `epi_df$geo_value` to `df$states` and `epi_df$county` to `df$county`.
```
dajmcdon, line 12: We should say what `NULL` the default does.

R/step_population_scaling.R
```diff
@@ -0,0 +1,193 @@
+#' Create a recipe step that scales variables using population data
+#'
+#' `step_population_scaling` creates a specification of a recipe step
+#' that will add a population scaled column in the data. For example,
+#' load a dataset that contains county population, and join to an `epi_df`
+#' that currently only contains number of new cases by county. Once scaled,
+#' predictions can be made on case rate.
```
dajmcdon, line 7: I think we should mention that there's nothing special about "population" (just that that's the most natural/common use case). It could scale by any variable.

R/step_population_scaling.R
```diff
@@ -0,0 +1,193 @@
...
+#' operations for this recipe. The recipe should contain information about the
+#' `epi_df` such as column names.
+#' @param df a data frame that contains the population data used for scaling.
+##' @param by A character vector of variables to join by. Column names should be
+#' in lower case.
+#' To join by different variables on the `epi_df` and `df`, use a named vector.
+#' For example, by = c("geo_value" = "states") will match `epi_df$geo_value`
+#' to `df$states`. To join by multiple variables, use a vector with length > 1.
+#' For example, by = c("geo_value" = "states", "county" = "county") will match
+#' `epi_df$geo_value` to `df$states` and `epi_df$county` to `df$county`.
```
dajmcdon, line 19: Same as above. Maybe also point to the `dplyr::full_join()` documentation.

R/step_population_scaling.R
```diff
@@ -0,0 +1,193 @@
...
+#'
+#' @param df_pop_col the name of the column in the data frame `df` that
+#' contains the population data and will be used for scaling.
+#' This should be one column, and column names should be in lower case.
+#' @param ... the corresponding column(s) in the `epi_df` that will be scaled.
+#' @param inputs Quosure(s) of `...`.
```
dajmcdon, line 25: I think you should rename this `columns`. And see [here](https://github.com/tidymodels/recipes/blob/HEAD/R/lag.R) to try and match args.

R/layer_population_scaling.R
```diff
@@ -0,0 +1,133 @@
...
+                                               object$df,
+                                               by= tolower(object$by)) %>%
+      dplyr::mutate(dplyr::across(dplyr::all_of(col_names),
+                                  ~.x * !!pop_col ,
+                                  .names = "{.col}{suffix}")) %>%
+      dplyr::select(- !!pop_col)
+
+    components
+
+
```
dajmcdon, line 132: let's delete some of the extra lines

tests/testthat/test-population_scaling.R
```diff
@@ -0,0 +1,164 @@
...
+                   time_value = rep(as.Date("2021-01-01") + 0:4, 2),
+                   case = 1:10,
+                   death = 1:10) %>%
+    epiprocess::as_epi_df()
+
+  r <-epi_recipe(newdata) %>%
+    step_population_scaling(df = pop_data,
+                            df_pop_col = "value", by = c("geo_value" = "states", "county" = "counties"),
+                            c("case", "death"), suffix = "_rate")
+
```
dajmcdon, line 49: what happens if the `by` selectors don't match? It should error out inside the `dplyr::full_join(...)`, but is that message interpretable? We may want to throw a more interpretable error earlier.

dajmcdon (changes requested):
Nice job. A few minor questions/comments, but this looks really good.

R/layer_population_scaling.R
```diff
@@ -0,0 +1,133 @@
+#' Revert population scaled prediction
+#'
+#' @param frosting a `frosting` postprocessor
+#' @param ... the column(s) in the `epi_df` to scale back.
+#' @param df a data frame that contains the population data used for scaling.
+#' @param by A character vector of variables to join by. Column names should be
+#' in lower case.
```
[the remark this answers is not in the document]
ChloeYou: Yah, that's true. I'll change this description. The upper/lowercase consistency is adjusted for in the functions so that the `join` can be performed smoothly. I have a unit test that checks this functionality.

R/layer_population_scaling.R
ChloeYou: added description and also unit tests for the default case.

R/step_population_scaling.R
ChloeYou: Good idea, I'll point to the the `dplyr::*_join()` document. I deliberately chose to use a `dplyr::left_join()` to avoid hassles of bringing in columns in the population data. Although we can always use `recipes::step_rm()` to remove the unwanted columns in the population data so it doesn't get mixed up in the modelling steps. I think I'll also make it clear in the documentation that a left join was performed.

tests/testthat/test-population_scaling.R
ChloeYou: When you say they don't match, do you mean it should report a message like this?
```
 t <- try(dplyr::left_join(new_data, object$df,
                            by= tolower(object$by)),
                silent = TRUE)
  if (grepl("Join columns must be present in data", t)){
      stop("columns in `by` selectors of `step_population_scaling`
           must be present in data and match")}
```

R/layer_population_scaling.R
```diff
@@ -0,0 +1,133 @@
+#' Revert population scaled prediction
+#'
+#' @param frosting a `frosting` postprocessor
+#' @param ... the column(s) in the `epi_df` to scale back.
```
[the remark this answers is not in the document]
ChloeYou: Is it just for the `...` parameter?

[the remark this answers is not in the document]
dajmcdon: Yes, that's right.

R/step_population_scaling.R
dajmcdon: Great, thanks!

R/layer_population_scaling.R
```diff
@@ -0,0 +1,149 @@
...
+  function(object, components, the_fit, the_recipe, ...) {
+    stopifnot("Only one population column allowed for scaling" =
+                length(object$df_pop_col) == 1)
+
+    t <- try(dplyr::left_join(components$predictions, object$df,
+                              by= tolower(object$by)),
+             silent = TRUE)
+    if (any(grepl("Join columns must be present in data", unlist(t)))){
+      stop("columns in `by` selectors of `layer_population_scaling` must be present in data and match")}
+
```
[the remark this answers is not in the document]
ChloeYou: So in the `bake` function, I intended to use the `try()` function on the left join to see if the `by` selector functions properly, and stop/throw an error accordingly. The actual left join is then performed later in the `bake` function. I used this method  instead of the usual `ifelse` check because of how the `by` selector are written. Sometimes it's written as `c("a" = "b")` or some other formats which is hard to assess. 

What do you think of the this check? Does is match what you're thinking of?

dajmcdon (approved):

PR #116, opened 2022-08-03
Intro vignette

dajmcdon:
In making the vignette, I (unfortunately) had to make small adjustments to lots of other things. So this is going to be a big pain. For that reason, I'm tagging everyone as a reviewer.


musings/.gitignore
```diff
@@ -0,0 +1 @@
+*.html
```
kenmawer, line 1: Was this addition of `gitignore` accidental or intentional?

vignettes/epipredict.Rmd
```diff
@@ -0,0 +1,360 @@
...
+# remotes::install_github("cmu-delphi/epipredict")
+library(epipredict)
+```
+
+
+# Goals for the package
+
+At a high level, our goal with `epipredict` is to make running simple Machine Learning / Statistical forecasters for epidemiology easy. However, this package is extremely extensible, and that is part of the utility. Our hope is that it is easy for users with epi training and some statistics to fit baseline models while still allowing those with more nuanced statistical understanding to create complicated specializations within the same framework.
+
+Serving both populations is the main motivation for our efforts, but at the same time, we have tried hard to make it useful.
```
kenmawer, line 33: What is the "both populations" you are speaking of?

vignettes/epipredict.Rmd
```diff
@@ -0,0 +1,360 @@
...
+preprocessing, training, and prediction, bound together, through a package called
+`{workflows}`. We built `{epipredict}` on top of that setup. In this way, you CAN
+use almost everything they provide.
+
+* However, `{workflows}` doesn't do postprocessing. And nothing in the -verse handles _panel data_.
+
+* The tidy-team doesn't have plans to do either of these things. (We checked).
+
+* There are two packages that do _time series_ built on `{tidymodels}`, but it's
+"basic" time series: 1-step AR models, exponential smoothing, STL decomposition, etc.[^2] We have never found these models to be particularly helpful for epidemic forecasting, but one could also integrate these methods into our framework.
```
kenmawer, line 93: What about `fable` and its related functions such as `autoplot`?a

vignettes/epipredict.Rmd
dajmcdon: Those mentioned in the previous paragraph (epi training + some stats vs. more nuanced statistical understanding). Not clear enough?

kenmawer: Alright, now I understand.

R/layer_residual_quantiles.R
```diff
@@ -68,12 +73,17 @@ slather.layer_residual_quantiles <-
     if (is.null(object$probs)) return(components)
 
     s <- ifelse(object$symmetrize, -1, NA)
-    r <- grab_residuals(the_fit, components)
-    q <- quantile(c(r, s * r), probs = object$probs, na.rm = TRUE)
+    r <- dplyr::bind_cols(
+      r = grab_residuals(the_fit, components),
+      geo_value = components$mold$extras$roles$geo_value,
+      components$mold$extras$roles$key) %>%
+      dplyr::group_by(dplyr::across(-r)) %>%
```
ChloeYou, line 80: Maybe I'm missing something in the code but does this grouping need to evaluate whether `object$by_key` is `TRUE` or `FALSE` before performing the group by?

R/arx_forecaster.R
```diff
@@ -83,28 +87,35 @@ arx_forecaster <- function(x, y, key_vars, time_value,
 #' arx_args_list()
 #' arx_args_list(symmetrize = FALSE)
 #' arx_args_list(levels = c(.1, .3, .7, .9), min_train_window = 120)
-arx_args_list <- function(lags = c(0, 7, 14), ahead = 7, min_train_window = 20,
-                          levels = c(0.05, 0.95), intercept = TRUE,
+arx_args_list <- function(lags = c(0L, 7L, 14L),
```
ChloeYou, line 90: Is there a reason why `intercept` was removed? Is it because it seems like we would always have the intercept in the ARX model no matter wha?

R/arx_forecaster_mod.R
```diff
@@ -1,29 +1,83 @@
...
+#'
+#' @param epi_data An `epi_df` object
+#' @param outcome A character (scalar) specifying the outcome (in the
+#'   `epi_df`).
+#' @param predictors A character vector giving column(s) of predictor
+#'   variables.
+#' @param trainer A `{parsnip}` model describing the type of estimation.
+#'   For now, we enforce `mode = "regression"`.
+#' @param args_list A list of customization arguments to determine
+#'   the type of forecasting model. See [arx_args_list()].
```
ChloeYou, line 16: Should users know the default of `arx_args_list()` creates 0, 7, 14 periods lagged predictors and 7 day ahead outcomes?

R/arx_forecaster_mod.R
```diff
@@ -1,29 +1,83 @@
...
+    r <- step_epi_lag(r, !!p, lag = lags[[l]])
+  }
+  r <- r %>%
+    step_epi_ahead(dplyr::all_of(!!outcome), ahead = args_list$ahead) %>%
+    step_epi_naomit()
   # should limit the training window here (in an open PR)
   # What to do if insufficient training data? Add issue.
-  # remove intercept? not sure how this is implemented in tidymodels
+
+  forecast_date <- args_list$forecast_date %||% max(epi_data$time_value)
```
ChloeYou, line 51: Trying to replicate this code... what does `%||%` mean? `||` means `or`, but what does the `%%` do?

R/zzz.R
```diff
@@ -0,0 +1,9 @@
+# ON LOAD ----
+
+# The functions below define the model information. These access the model
+# environment inside of parsnip so they have to be executed once parsnip has
+# been loaded.
+
+.onLoad <- function(libname, pkgname) {
```
ChloeYou, line 7: What is this file for? It automatically loads the `make_flatline_reg()` function as a `parsnip` model after the `epipredict` package is loaded?

ChloeYou (commented):
Left some comments asking about how some functions work. Might save time to address them in-person during a group meeting. Thank you!

R/arx_forecaster.R
```diff
@@ -34,7 +34,7 @@ arx_forecaster <- function(x, y, key_vars, time_value,
   }
 
   dat <- create_lags_and_leads(x, y, lags, ahead, time_value, keys)
-  if (intercept) dat$x0 <- 1
```
mgyliu, line 37: OOC, where was this intercept coming from before? I don't see it passed in or defined above

rachlobay: Based on last week's meeting, I think that is obsolete.

R/get_test_data.R
```diff
@@ -21,22 +21,19 @@
...
+  max_lags <- max(map_dbl(recipe$steps, ~ max(.x$lag %||% 0)), 0)
 
   # CHECK: Return NA if insufficient training data
   if (dplyr::n_distinct(x$time_value) < max_lags) {
     cli_stop("You supplied insufficient training data for this recipe. ",
              "You need at least {max_lags} distinct time_values.")
   }
 
-  groups <- epi_keys(recipe)[epi_keys(recipe) != "time_value"]
+  groups <- epi_keys(recipe)[-1]
```
mgyliu, line 36: It's not immediately clear to me what this line is trying to do. I think it would be helpful to have a comment here to explain "exclude time_value", or "epi_keys will return with time_value as the first element"

Side note, I totally thought `[-1]` was grabbing the last element of the vector until I looked it up

tests/testthat/test-epi_recipe.R
```diff
@@ -65,7 +65,7 @@ test_that("epi_recipe formula works", {
     variable = "z", type = "nominal", role = "key",
     source = "original")
 
-  expect_identical(r$var_info, ref_var_info)
+  #expect_identical(r$var_info, ref_var_info)
```
mgyliu, line 68: Did we want to leave this test in?

vignettes/epipredict.Rmd
```diff
@@ -0,0 +1,360 @@
...
+All the forcasters we provide are built on our framework. So we will use these basic models to illustrate its flexibility.
+
+## Forecasting framework
+
+Our framework for creating custom forecasters views the prediction task as a set of modular components. There are four types of components:
+    
+1. Preprocessor: do things to the data before model training
+2. Trainer: train a model on data, resulting in a fitted model object
+3. Predictor: make predictions, using a fitted model object and processed test data
+4. Postprocessor: do things to the predictions before returning
```
mgyliu, line 56: Nitpick but could we use language a bit more precise than "do things"? Maybe "manipulate data/predictions"?

R/arx_forecaster_mod.R
mgyliu: I had the same question. From what I understand, `x %||% y` is the same as `if (x is NULL) {return y} else {return x}`

R/arx_forecaster_mod.R
```diff
@@ -1,29 +1,83 @@
...
+  target_date <- args_list$target_date %||% forecast_date + args_list$ahead
   f <- frosting() %>%
     layer_predict() %>%
-    layer_naomit(.pred) %>%
-    layer_residual_quantile(
```
mgyliu, line 18: Did this used to be a function or was this a typo before?

R/epi_keys.R
```diff
@@ -29,7 +29,7 @@ epi_keys.recipe <- function(x) {
 epi_keys_mold <- function(mold) {
   keys <- c("time_value", "geo_value", "key")
   molded_names <- names(mold$extras$roles)
-  mold_keys <- purrr::map_chr(mold$extras$roles[molded_names %in% keys], names)
+  mold_keys <- map_chr(mold$extras$roles[molded_names %in% keys], names)
```
mgyliu, line 32: Will this work when there are multiple `other_key`s? I know we tried it in the meeting last week but I can't remember. The fix I had was to just use `map` and then `unlist(unname(mold_keys))`.

R/zzz.R
mgyliu: +1 I am also curious

DESCRIPTION
```diff
@@ -32,7 +32,6 @@ Imports:
     glue,
     hardhat (>= 1.2.0),
     magrittr,
-    purrr,
```
mgyliu, line 35: Is this just being removed b/c we implemented our own versions of the `purrr` functions?

R/arx_forecaster.R
mgyliu: Should we necessarily have these default `lags` and `ahead` arguments? The defaults seem specifically defined for `time_type` = `day` but there are many other time types for which the defaults make a bit less sense

rachlobay: If we do leave the defaults as is, it may be good to briefly mention the defaults somewhere else in `arx_args_list()` aside from the initial list of arguments to emphasize them (under the argument descriptions seems to be natural choice, where I see mention of defaults for some but not all of the arguments).

R/arx_forecaster_mod.R
rachlobay: I think that was the previous function name for `layer_residual_quantiles`, so it should've been updated to `layer_residual_quantiles`.

R/flatline_epi_forecaster.R
```diff
@@ -0,0 +1,116 @@
...
+  validate_forecaster_inputs(epi_data, outcome, "time_value")
+  keys <- epi_keys(epi_data)
+  ek <- keys[-1]
+  outcome <- rlang::sym(outcome)
+
+
+  r <- epi_recipe(epi_data) %>%
+    step_epi_ahead(!!outcome, ahead = args_list$ahead, skip = TRUE) %>%
+    recipes::update_role(!!outcome, new_role = "predictor") %>%
+    recipes::add_role(dplyr::all_of(keys), new_role = "predictor")
```
rachlobay, line 43: I see these functions, `update_role()` and `add_role()` are also used where the flatline forecaster is demonstrated at the end of the getting started vignette... Since they seem quite useful, I think their use in preprocessing could be mentioned in a line where that is most appropriate to fit.

R/get_test_data.R
rachlobay: The assumption that `time_value` is the 1st element is fair at this time, but it's probably best to just use the col name there if possible because if at some point in the future we choose not to enforce col. ordering, this could be problem.

R/get_test_data.R
```diff
@@ -21,22 +21,19 @@
 #'  get_test_data(recipe = rec, x = case_death_rate_subset)
 #' @export
 
-get_test_data <- function(recipe, x){
-  # TO-DO: SOME CHECKS OF THE DATASET
-  if (any(!(c('geo_value','time_value') %in% colnames(x)))) {
-    rlang::abort("`geo_value`, `time_value` does not exist in data")
-  }
-  ## CHECK if it is epi_df?
```
rachlobay, line 29: Guess this is covered by the `is.data.frame()` check below?

R/zzz.R
rachlobay: I am as well... So when the parsnip package is loaded, the code `make_flatline_reg()` is run b/c it needs access some enviro in there. The package stores information about the models in an internal environment object, as it says [here](https://www.tidymodels.org/learn/develop/models/), so we must read and write to that enviro to set and get model-related stuff?

R/layer_residual_quantiles.R
dajmcdon: No you're not missing anything. Current implementation ignores that option and assumes it's TRUE. I haven't adjusted it to handle the FALSE case yet.

R/arx_forecaster.R
dajmcdon: On Intercept, yes, it's always there.

On the `time_type`. Good point. Our implementation always modifies the `time_value` by a day. 

Actually, can someone create an issue to:
1. adjust the lead/lag function to accept a `time_unit` argument (days, months, weeks, years) and perform the operations appropriately.
2. Add an arg here (and elsewhere) to allow a user to specify the units (possibly different for lags/ahead/train_window)

R/arx_forecaster_mod.R
dajmcdon: I agree, but I don't want to put this in the docs at the moment. Looking at the function would tell you.

R/arx_forecaster_mod.R
dajmcdon: Yup!

R/zzz.R
dajmcdon: Yes, this is right. It took me a while of reading through those docs to get it right.

So, when we write quantile regression, the same stuff applies.

R/get_test_data.R
dajmcdon: Agreed. Fixing.

vignettes/epipredict.Rmd
dajmcdon: Oh, fine...

R/arx_forecaster_mod.R
dajmcdon: Right, and I actually DON'T want to omit any NA's

dajmcdon: But leaving for historical reasons (and in case there's an example of where I'd want it).

R/epi_keys.R
dajmcdon: It should. All `other_keys` get assigned the role "key". So this is a place where screwing up the `epi_df` metadata could break epipredict.

DESCRIPTION
dajmcdon: Yup. `purrr` is a huge package. So better to just keep the ones we need directly.

R/flatline_epi_forecaster.R
dajmcdon: In think @ChloeYou is doing a Vignette that talks about recipes/tidymodels more. Can you mention there?

R/get_test_data.R
dajmcdon: There's some weirdness that can happen here. The recipe template should have all the keys and any data needed to prep/bake the test data for the predictions. So we want to be sure that the new data has all of that. If not, this should fail.

(In other words, this has been replaced with a more stringent requirement.)

R/layer_residual_quantiles.R
dajmcdon: I've addressed this below. The updates should now
* No grouping by default.
* Allow grouping using a character vector of keys, as many as desired.

R/epi_keys.R
mgyliu: I pulled your branch and ran this test that I wrote for `test-epi_keys`, and it doesn't pass. Maybe I'm missing something? @dajmcdon 

```
test_that("epi_keys_mold extracts additional keys when they are present", {
  my_data <- case_death_rate_subset %>%
    sample_n(6) %>%
    tsibble::as_tsibble() %>% # add 2 extra keys
    mutate(state = rep("MA", 6), pol = rep("blue", 6)) %>% 
    as_epi_df(additional_metadata = list(other_keys=c("state", "pol"))) # <-- metadata constructed here

  my_recipe <- epi_recipe(my_data) %>%
    step_epi_ahead(death_rate , ahead = 7) %>%
    step_epi_naomit()

  my_workflow <- epi_workflow(my_recipe, linear_reg()) %>% fit(my_data)

  expect_equal(
    epi_keys_mold(my_workflow$pre$mold), 
    c("time_value", "geo_value", "state", "pol"))
})
```

Error msg:
```
Error in `vapply(.x, .f, .mold, ..., USE.NAMES = FALSE)`: values must be length 1,
 but FUN(X[[3]]) result is length 2
Backtrace:
 1. testthat::expect_equal(...)
 4. epipredict::epi_keys_mold(my_workflow$pre$mold)
 5. epipredict::map_chr(...)
      at epipredict/R/epi_keys.R:32:2
 6. epipredict::.rlang_purrr_map_mold(.x, .f, character(1), ...)
      at epipredict/R/compat-purrr.R:32:2
 7. base::vapply(.x, .f, .mold, ..., USE.NAMES = FALSE)
      at epipredict/R/compat-purrr.R:49:2

Error in reporter$stop_if_needed() : Test failed
Backtrace:
1: stop(error)
2: (function (envir) 
3: test_that("epi_keys_mold extracts additional keys when they are present", 
```

dajmcdon: Nuts. Good catch. Should be fixed now. I added your code above to the test.

mgyliu (approved):
lgtm!

R/flatline_epi_forecaster.R
ChloeYou: Noted! Will add to the vignette.

R/epi_keys.R
dajmcdon: Bugger. This fails tests. The reason being that `tsibble` is not imported. I don't want it for one test. And I'm having trouble adding the additional metadata without converting to tsibble. Can you fix it up on your local version?

mgyliu: Here, this one doesn't use `tsibble`:
```
test_that("epi_keys_mold extracts additional keys when they are present", {
  my_data <- tibble::tibble(
    geo_value = rep(c("ca", "fl", "pa"), each = 3), 
    time_value = rep(seq(as.Date("2020-06-01"), as.Date("2020-06-03"),
                            by = "day"), length.out = length(geo_value)),
    pol = rep(c("blue", "swing", "swing"), each = 3), # extra key
    state = rep(c("ca", "fl", "pa"), each = 3), # extra key
    value = 1:length(geo_value) + 0.01 * rnorm(length(geo_value))
  ) %>%
  as_epi_df(additional_metadata = list(other_keys = c("state", "pol")))

  my_recipe <- epi_recipe(my_data) %>%
    step_epi_ahead(value , ahead = 7) %>%
    step_epi_naomit()

  my_workflow <- epi_workflow(my_recipe, linear_reg()) %>% fit(my_data)

  expect_setequal(
    epi_keys_mold(my_workflow$pre$mold), 
    c("time_value", "geo_value", "state", "pol"))
})
```

mgyliu: Did you want me to push it to your branch myself? I can definitely do that too

R/layer_residual_quantiles.R
```diff
@@ -34,46 +36,79 @@
...
     if (is.null(object$probs)) return(components)
 
+
     s <- ifelse(object$symmetrize, -1, NA)
-    r <- grab_residuals(the_fit, components)
-    q <- quantile(c(r, s * r), probs = object$probs, na.rm = TRUE)
+    r <- dplyr::bind_cols(
+      r = grab_residuals(the_fit, components),
+      geo_value = components$mold$extras$roles$geo_value,
+      components$mold$extras$roles$key)
```
kenmawer, line 88: What is the column name for this? I know there are `r` and `geo_value` column names.

dajmcdon: `components$mold$extras$roles$key` can be 2 things.
1. It will be `NULL` if there are no additional keys. In this case, `r` will contain 2 columns `r` and `geo_value`
2. It will be a data.frame with one column for each key. Then this data frame will be bound as is to the other two columns, preserving the column names.

dajmcdon:
Closes #117

PR #119, opened 2022-08-05
Vignette for visualizations (first draft)

kenmawer:
This is a rough draft. I put in information that I believe would be relevant, although it may have already been reiterated from another vignettes that illustrates a similar feature. This is far from perfect, but just meant to get an idea of what things should or shouldn't be discussed when discussing about visualizations to a user.


dajmcdon (commented):
@kenmawer Looking over this, I think it's a good start, but perhaps not quite the right focus. The emphasis should really be on "using visualizations to understand predictions".  Unfortunately, this vignette comes after the others, so it's not that easy to do "right now".

Some ideas:
1. So it may actually be good to branch off `draft-intro` and build from there. Use the predictions and the workflow and try to investigate the model fit.
2. Use some of the sliding tools in epiprocess to produce forcasts at multiple dates. 
3. Look at some of the things we did with [`evalcast`](https://cmu-delphi.github.io/covidcast/evalcastR/articles/evalcast.html) or [`covidHubUtils`](http://reichlab.io/covidHubUtils/articles/covidHubUtils-overview.html) (the fan plots there are especially nice). Or some trajectories like [these](https://htmlpreview.github.io/?https://github.com/cmu-delphi/covid-19-forecast/blob/main/forecasts/state-output/anteater_2021-07-19.html?raw=true)

dajmcdon:
Possibly superseded by #121

Issue #110, opened 2022-07-26
Vignette - Demonstration of sliding


rachlobay:
Demonstration of sliding using relevant functions from epipredict... Probably base this on the vignette for epiprocess & the simpleforecasts.Rmd musing. Note how the functions from epipredict may complicate things.

rachlobay:
Addressed by PR #125.

PR #125, opened 2022-08-15
Vignette - demonstration of sliding

rachlobay:
Addresses Issue #110. 

Currently has 
- Two ex. of sliding with diff. engines
- One ex. of version-aware and unaware forecasting


vignettes/sliding.Rmd
```diff
@@ -0,0 +1,281 @@
...
+#library(covidcast)
+library(data.table)
+library(dplyr)
+library(tidyr)
+library(ggplot2)
+```
+
+# Demonstrations of sliding AR and ARX forecasters
+
+An key function from the epiprocess package is `epi_slide()`, which allows the user to apply a function or formula-based computation an `epi_df` over a running window of `n` time steps that may or may not be grouped (see [`epiprocess` "Slide a computation over signal values" vignette](https://cmu-delphi.github.io/epiprocess/articles/advanced.html#version-aware-forecasting-revisited-1) to go over the basics of this function). The equivalent sliding method for an `epi_archive` object can be called by using the wrapper function `epix_slide()` The key difference from `epi_slide()` is that it performs version-aware computations. That is, the function only uses data that would have been available as of time t, for that reference time.
```
dajmcdon, line 31: "A key function..."

* Maybe also mention why this is useful here? (backtesting our forecaster)

vignettes/sliding.Rmd
```diff
@@ -0,0 +1,281 @@
...
+```{r make-arx-kweek, warning = FALSE}
+# Latest snapshot of data, and forecast dates
+x_latest <- epix_as_of(x, max_version = max(x$versions_end)) 
+fc_time_values <- seq(as.Date("2020-08-01"), as.Date("2021-12-01"), 
+                      by = "1 month")
+
+k_week_ahead <- function(ahead = 7, engine_type = "linear_reg") {
+  if(engine_type == "linear_reg"){
+    x_latest %>%
+      epi_slide(function(x, ...)
```
dajmcdon, line 80: Does this work with 

```r
x_latest %>% 
  epi_slide(~arx_forecaster(.x, outcome=...)
```

vignettes/sliding.Rmd
```diff
@@ -0,0 +1,281 @@
...
+  if(engine_type == "linear_reg"){
+    x_latest %>%
+      epi_slide(function(x, ...) 
+        arx_forecaster(x, outcome = "case_rate", 
+                       predictors = c("case_rate", "percent_cli"),  
+                       args_list = arx_args_list(ahead = ahead))$predictions %>% 
+          select(-c(geo_value, time_value)),
+        n = 120, ref_time_values = fc_time_values, new_col_name = "fc") %>% 
+      mutate(engine_type = "linear_reg")
+  } else if (engine_type == "random_forest"){
```
dajmcdon, line 87: Other than the engine, everything else is the same right? Can we simplify this function a bit? Maybe set
```r
engine = ifelse(engine_type = "linear_reg", linear_reg, rand_forest(mode="regression"))
```
and just pass `engine` as the argument?

vignettes/sliding.Rmd
```diff
@@ -0,0 +1,281 @@
...
+                       args_list = arx_args_list(ahead = ahead))$predictions %>% 
+          select(-c(geo_value, time_value)),
+        n = 120, ref_time_values = fc_time_values, new_col_name = "fc") %>% 
+      mutate(engine_type = "random_forest")
+  } 
+}
+
+# Generate the forecasts, and bind them together
+fc <- bind_rows(
+  purrr::map_dfr(c(7,14,21,28), ~ k_week_ahead(.x, engine_type = "linear_reg")),
```
dajmcdon, line 102: `map_dfr` is in the `epipredict` NAMESPACE, so I don't think you need the `purrr` version.

vignettes/sliding.Rmd
```diff
@@ -0,0 +1,281 @@
...
+              package = "epipredict", mustWork = TRUE)
+  ) %>%
+  group_by(version, geo_value) %>% 
+  arrange(time_value) %>% 
+  mutate(cr_7dav = RcppRoll::roll_meanr(case_rate, n = 7L)) #%>%
+  #filter(geo_value %in% c('Alberta', "BC"))
+can <- as_epi_archive(can)
+can_latest <- epix_as_of(can, max_version = max(can$DT$version))
+
+can_k_week_ahead <- function(ahead = 7, engine_type = "linear_reg") {
```
dajmcdon, line 151: Is there anyway to merge this function with the other one? Create 1 single function that can be used in both cases?

vignettes/sliding.Rmd
```diff
@@ -0,0 +1,281 @@
...
+  geom_line(aes(y = fc_.pred)) + geom_point(aes(y = fc_.pred), size = 0.5) +
+  geom_vline(aes(xintercept = time_value), linetype = 2, alpha = 0.5) +
+  facet_wrap(~ geo_value, scales = "free_y", ncol = 3) +
+  scale_x_date(minor_breaks = "month", date_labels = "%b %y") +
+  labs(title = "Using boosted regression trees", x = "Date", 
+       y = "Reported COVID-19 case rates") +
+  theme(legend.position = "none")  
+```
+
+Both approaches tend to produce quite volatile forecasts (point predictions) and/or are overly confident (very narrow bands), particular when boosted regression trees were used. As this is meant to be a simple demonstration of sliding with different engines in `arx_forecaster`, we may devote another vignette to work on improving the predictive modelling using the suite of tools available in epipredict.
```
dajmcdon, line 220: "particularly"

dajmcdon (commented):
Mostly minor comments.

One thing that occurs to me: in the US example at the beginning, can you fit on all states and territories and just display 2 states? I'd rather demonstrate that the forecasts are at least reasonable, so the extra data should help.

Maybe that means adding another signal to the Canadian data? I guess it's not obvious that there are any.

vignettes/sliding.Rmd
rachlobay: Yes. Sliding with a formula or function format works... I'll use the formula format if that's preferred.

rachlobay:
For the US example at the beginning, I was able to fit on all states, but not the territories as I was getting an error that looked to be due to their incomplete data. Then, I displayed & briefly discussed the results for CA and FL as before.

vignettes/sliding.Rmd
rachlobay: I don't see that fun. in the `epipredict` NAMESPACE yet?

vignettes/sliding.Rmd
dajmcdon: Yeah, It looks a bit cleaner.

vignettes/sliding.Rmd
dajmcdon: You're right. It's internal and not exported. NM.

vignettes/sliding.Rmd
dajmcdon: There's still a few typos in this sentence.

vignettes/sliding.Rmd
```diff
@@ -74,33 +75,28 @@ x_latest <- epix_as_of(x, max_version = max(x$versions_end))
...
-  } 
+k_week_ahead <- function(epi_df, outcome, predictors, ahead = 7, engine) {
+  epi_df %>%
+    epi_slide(~arx_forecaster(.x, outcome = outcome, 
+                              predictors = predictors,  
+                              trainer = eval(parse(text = engine)), 
+                              args_list = arx_args_list(ahead = ahead))$predictions %>% 
+                select(-c(geo_value, time_value)), n = 120, 
+              ref_time_values = fc_time_values, new_col_name = "fc") %>% 
+    mutate(engine_type = ifelse(engine == "linear_reg()", "linear_reg",
```
dajmcdon, line 86: Try `case_when()` instead of the nested `if_else()`.

dajmcdon (changes requested):

vignettes/sliding.Rmd
rachlobay: Should be fixed now.

dajmcdon (approved):

PR #130, opened 2022-08-26
126 check for dates

### LINKED ISSUE #126: `args_list` `forecast_date` and `target_date`: assertion for `Date` type
While using `arx_args_list`, wasn't immediately clear to me that `forecast_date` and `target_date` have to be `Date` types, and won't work correctly if it's a string like "2022-08-08". If it's not a `Date`, it gives this warning
```
Warning message:
Incompatible methods ("<.vctrs_vctr", "<.Date") for "<" 
```
which comes from [`layer_add_forecast_date.R#L85`](https://github.com/cmu-delphi/epipredict/blob/172ab1e8352d7d16e8b9f5c581f3ccd3c921734c/R/layer_add_forecast_date.R#L85) 

Proposed change is to either
- update documentation to make argument type known and clear
- change code to throw a readable error earlier if the user gives a non-`Date`
- or both?

dajmcdon:
Closes #126.

Added a check in both functions.
Also modified the documentation to specify `Date.`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -64,5 +64,5 @@ Remotes:
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.0
+RoxygenNote: 7.2.1
 LazyData: true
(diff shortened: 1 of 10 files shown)


mgyliu (changes requested):
Thanks for making this PR! 

Based on these changes, my understanding is that other calls to `layer_add_forecast_date` with a `string` or other non-`Date` will still error out at [this line](https://github.com/cmu-delphi/epipredict/blob/frosting/R/layer_add_forecast_date.R#L85). Should we maybe be checking the date format in there instead? `arx_forecaster` and `flatline_forecaster` both have calls to `layer_add_forecast_date` too.

dajmcdon:
Good call. How's this look?

mgyliu (approved):
lgtm 🚀

PR #131, opened 2022-08-30
105 parameter tuning

ChloeYou:
This is a demo vignette to show how I approached cross validation and hyperparameter tuning, so it doesn't need to be merged into the main branch. The current Epipredict workflow doesn't seem to directly support CV and parameter tuning, good news is that they can still be achieved with a few extra lines of code. The biggest issue seems to come from the recipes step - the model doesn't recognize the new variables created in recipes and causes some issues with subsequent steps in CV.  The way to work around it is it prep and bake the data before passing it into the workflow for CV & tuning.

@dajmcdon Would you be able to take a look when you're available? Let me know if you want to discuss more on it. Thank you!


dajmcdon (approved):
test comment

PR #134, opened 2022-09-14
133 fix residuals

### LINKED ISSUE #133: Residuals from non-lm engines
Looking at the `sliding.Rmd` vignette, the random forest prediction intervals are very narrow. Is there a bug somewhere in the `residual_quantiles()`?

dajmcdon:
Closes #133.

The basic issue is that the test data must be complete (or at least have enough non-NAs to do something). In the sliding vignette, many locations are missing recent `percent_cli` so the predictions got a bit odd.

Changes:
* Update to `get_test_data()` to allow for NA fill and print messages if there's insufficient data for the fill.
* Move `epiprocess` to Depends, since this package is fundamentally linked to it.
* Changed default `residual_quantile` values (the old ones were strange).


rachlobay (approved):
After fixing that one test & adding a little more description to `n_recent` in `get_test_data`, looks good to merge (after all checks pass).

PR #182, opened 2023-05-23
Update to latest `epidatr` (`fetch_tbl` -> `fetch`)

brookslogan:
Resolves breaking change from upstream cmu-delphi/epidatr#99.  We "can't" `>=` the epidatr version dependency to try to get people better error messages because old versions hanging around are tagged "1.0.0" or maybe also "0.5.0", and we don't want to call the new version with this change something > "1.0.0".

### DIFF
--- a/data-raw/case_death_rate_subset.R
+++ b/data-raw/case_death_rate_subset.R
@@ -11,7 +11,7 @@ x <- covidcast(
   time_values = epirange(20201231, 20211231),
   geo_values = "*"
 ) %>%
-  fetch_tbl() %>%
+  fetch() %>%
   select(geo_value, time_value, case_rate = value)
 
 y <- covidcast(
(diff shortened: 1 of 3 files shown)


dajmcdon:
Just to check, this `epidatr` is now the version on `main`?

brookslogan:
Yes.  I have another PR in epidatr pending that will bump the version to 0.6.0, though I'm not sure if it's useful for epipredict due to the old-1.0.0 problem.

dajmcdon:
Ah, I see. That's the `epidatr@dev` but epipredict uses `epidatr@main` currently. So we'll need to update the DESCRIPTION here as well.

brookslogan:
I don't think we need to update epipredict's DESCRIPTION.  It doesn't specify `@main` so it will use the default branch, which is now `@dev`.

dajmcdon (approved):

Issue #199, opened 2023-06-15
`get_test_data()` returns too many rows in some cases


dajmcdon:
For example, we should see only 1 row in the baked data below.

``` r
library(epipredict)
ca <- case_death_rate_subset %>%
  filter(geo_value == "ca")
  
rec <- epi_recipe(ca) %>%
  step_epi_lag(case_rate, lag = c(2, 4, 6)) %>%
  step_epi_ahead(case_rate, ahead = 7) %>%
  step_epi_naomit()

td <- get_test_data(rec, ca)

bake(prep(rec, ca), td)
#> # A tibble: 3 × 8
#>   time_value geo_value case_rate death_rate lag_2_case_rate lag_4_case_rate
#>   <date>     <chr>         <dbl>      <dbl>           <dbl>           <dbl>
#> 1 2021-12-31 ca             84.4      0.142            63.7            48.4
#> 2 2022-01-01 ca             NA       NA                76.0            54.9
#> 3 2022-01-02 ca             NA       NA                84.4            63.7
#> # ℹ 2 more variables: lag_6_case_rate <dbl>, ahead_7_case_rate <dbl>
```

<sup>Created on 2023-06-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The problem is because we didn't request `lag = 0`. 

Possible fix is to add some code to `get_test_data()` to remove the last two rows in `td`:

``` r
bake(prep(rec, ca), td[1:5, ])
#> # A tibble: 1 × 8
#>   time_value geo_value case_rate death_rate lag_2_case_rate lag_4_case_rate
#>   <date>     <chr>         <dbl>      <dbl>           <dbl>           <dbl>
#> 1 2021-12-31 ca               NA         NA            63.7            48.4
#> # ℹ 2 more variables: lag_6_case_rate <dbl>, ahead_7_case_rate <dbl>
```

<sup>Created on 2023-06-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This gives the correct results except, possibly, when we also use `step_growth_rate()`.

Issue #209, opened 2023-06-27
arx forecaster: first lag "duplicates" fits in bizzare ways


dsweber2:
when the first lags is larger than zero, the output is duplicated that many times +1
```r
jhu <- case_death_rate_subset
jhu <- jhu %>% filter(time_value >= max(time_value) - 30)
out0 <- arx_forecaster(
  jhu,
  outcome = "death_rate",
  predictors = c("case_rate", "death_rate"),
  args_list = arx_args_list(
  lags = c(0))
)
out1 <- arx_forecaster(
  jhu,
  outcome = "death_rate",
  predictors = c("case_rate", "death_rate"),
  args_list = arx_args_list(
  lags = c(1))
)
out2 <- arx_forecaster(
  jhu,
  outcome = "death_rate",
  predictors = c("case_rate", "death_rate"),
  args_list = arx_args_list(
  lags = c(2))
)
```

dajmcdon:
This is due to #207

rachlobay:
By output, you mean when we look at the preds? Such as `out2$predictions %>% filter(geo_value == "ak")` returning duplicate rows? If so, I think this may now be fixed with PR #211.

dsweber2:
Glad to hear it! Yes, that is what I mean, and it does sound like it's downstream of what you're working on. If I understand correctly, if you add something like "closes #209" to the first message in your pull request, you can get this issue to [auto-close](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) when that merges.

PR #211, opened 2023-07-03
Modified `get_test_data()` to omit excess rows when lag != 0

rachlobay:
Addresses Issue #199 (modified `get_test_data()` function + added a test). Basically, omitted the end rows according to the min # of lags in the recipe when that's not 0.

### DIFF
--- a/R/get_test_data.R
+++ b/R/get_test_data.R
@@ -43,6 +43,7 @@ get_test_data <- function(recipe, x, fill_locf = FALSE, n_recent = NULL) {
 
   if (!all(colnames(x) %in% colnames(recipe$template)))
     cli_stop("some variables used for training are not available in `x`.")
+  min_lags <- max(map_dbl(recipe$steps, ~ min(.x$lag %||% 0)), 0)
   max_lags <- max(map_dbl(recipe$steps, ~ max(.x$lag %||% 0)), 0)
   max_horizon <- max(map_dbl(recipe$steps, ~ max(.x$horizon %||% 0)), 0)
   min_required <- max_lags + max_horizon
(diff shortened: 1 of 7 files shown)


dajmcdon (changes requested):
I'm not yet convinced that this works. Can you add a test to make sure that we're getting the correct values? I'm actually not entirely sure what the right answer is here. Suppose I want to predict 3 days ahead with lags 2 and 4. My data are `1:10` with 10 being the most recent observation. Then I should use 8 and 6 as my `x` right? Is that what you get?

rachlobay:
Ok. We could map your ex. onto the current test & look at the results:

```
ca <- case_death_rate_subset %>%
  filter(geo_value == "ca") %>% 
  slice_tail(n = 10)

rec <- epi_recipe(ca) %>%
  step_epi_lag(case_rate, lag = c(2, 4)) %>% # Expect these to be 63.7, 48.4, respectively, according to the case_rate values in rows 8, 6 of ca
  step_epi_ahead(case_rate, ahead = 3) %>% 
  step_epi_naomit()

td <- get_test_data(rec, ca)

bake(prep(rec, ca), td)
```

Output:

```
# A tibble: 1 × 7
  time_value geo_value case_rate death_rate lag_2_case_rate lag_4_case_rate ahead_3_case_rate
  <date>     <chr>         <dbl>      <dbl>           <dbl>           <dbl>             <dbl>
1 2021-12-31 ca               NA         NA            63.7            48.4                NA
```

I've added checks to ensure that the lags match the values in the original data for a similar example... Is this along the lines of what you want me to check into or is there something else you had in mind?

dajmcdon:
Yes, exactly. But can you make it as simple as possible? Just create a minimal epi_df with obvious values that are easy to track.

rachlobay:
Added a simpler ex. to the get_test_data() tests based on the situation you suggested above.
Ex. Set-up:
```
  toy_epi_df <- tibble::tibble(
    time_value = seq(as.Date("2020-01-01"), by = 1,
                     length.out = 10),
    geo_value = "ak",
    x = 1:10
  ) %>% epiprocess::as_epi_df()

  toy_rec <- epi_recipe(toy_epi_df) %>%
    step_epi_lag(x, lag = c(2, 4)) %>%
    step_epi_ahead(x, ahead = 3) %>%
    step_epi_naomit()

```

rachlobay:
Also, this looks to address #209, so I added that to the development.

dajmcdon (approved):

Issue #213, opened 2023-07-04
Error when run `quantile_reg()` doc and tooling book examples in R 4.0.4


rachlobay:
When we try run the `quantile_reg()` doc ex in an older R version (I'll use 4.0.4 here), we get the following error:

```
tib <- data.frame(y = rnorm(100), x1 = rnorm(100), x2 = rnorm(100))
rq_spec <- quantile_reg(tau = c(.2, .8)) %>% set_engine("rq")
ff <- rq_spec %>% fit(y ~ ., data = tib)
predict(ff, new_data = tib)
Error in FUN(newX[, i], ...) : unused argument (simplify = FALSE)
Called from: apply(x, 1, function(q) unname(sort(q)), simplify = FALSE)
```

Also, the issue persists when we run the `quantile_reg` part from Section 8.6 of the tooling book:

```
jhu <- case_death_rate_subset %>%
  filter(time_value >= max(time_value) - 30)

out_gb <- arx_forecaster(
jhu, "death_rate", c("case_rate", "death_rate"),
quantile_reg()
)
Error in FUN(newX[, i], ...) : unused argument (simplify = FALSE)
Called from: apply(x, 1, function(q) unname(sort(q)), simplify = FALSE)
```

Looks like this is due to `apply(..., simplify = FALSE)` being introduced in a later R version, R 4.1, (a similar issue encountered in another package https://github.com/ModelOriented/shapviz/issues/29). 

The epipredict package currently depends on R (>= 3.5.0). So, we could either bump up the R version required for the package, or to make it compatible with lower versions, we could modify the offending line of code, which looks to be line 107 here: https://github.com/cmu-delphi/epipredict/blob/main/R/make_quantile_reg.R.

Now if we just remove the simplify function and try run the `quantile_reg()` doc ex, then I run into this error:
```
Error in `as_list_of()` at epipredict/R/dist_quantiles.R:55:2:
! Can't convert `..1` <double[,100]> to <double>.
Can't decrease dimensionality from 2 to 1.
Run `rlang::last_trace()` to see where the error occurred.
Called from: signal_abort(cnd, .file)
```
so the data is not agreeing well with the `dist_quantiles()` function.

What seems to do the trick to get the data in the required format to be used by `dist_quantiles()` is to first remove simplify and then wrap the apply function in `unname(as.list(as.data.frame()))`. This structure is based on the documentation ex provided for `dist_quantiles()` in which lists of a similar structure are inputted into that function.

PR #214, opened 2023-07-04
Possible fix so that people may be able to use lower versions of R

rachlobay:
Addresses #213 so that users may be able to use the epipredict package in lower versions of R. To do this, in `make_quantile_reg()` removed `simplify = FALSE` and wrapped apply in  `unname(as.list(as.data.frame()))`, so that we get a list that looks to be compatible with `dist_quantiles()` - similar in format to the documentation example given there.

### DIFF
--- a/R/make_quantile_reg.R
+++ b/R/make_quantile_reg.R
@@ -102,9 +102,9 @@ make_quantile_reg <- function() {
     # can't make a method because object is second
     out <- switch(
       type,
-      rq = dist_quantiles(x, object$tau), # one quantile
+      rq = dist_quantiles(unname(as.list(x)), object$tau), # one quantile
       rqs = {
-        x <- unname(apply(x, 1, function(q) unname(sort(q)), simplify = FALSE))
+        x <- unname(as.list(as.data.frame(apply(x, 1, function(q) unname(sort(q))))))
         dist_quantiles(x, list(object$tau))
       },
       rlang::abort(c("Prediction not implemented for this `rq` type.",
--- a/R/make_quantile_reg.R
+++ b/R/make_quantile_reg.R
@@ -104,7 +104,7 @@ make_quantile_reg <- function() {
       type,
       rq = dist_quantiles(unname(as.list(x)), object$tau), # one quantile
       rqs = {
-        x <- unname(as.list(as.data.frame(apply(x, 1, function(q) unname(sort(q))))))
+        x <- lapply(unname(split(x, seq(nrow(x)))), function(q) sort(q))
         dist_quantiles(x, list(object$tau))
       },
       rlang::abort(c("Prediction not implemented for this `rq` type.",


R/make_quantile_reg.R
```diff
@@ -102,9 +102,9 @@ make_quantile_reg <- function() {
     # can't make a method because object is second
     out <- switch(
       type,
-      rq = dist_quantiles(x, object$tau), # one quantile
+      rq = dist_quantiles(unname(as.list(x)), object$tau), # one quantile
       rqs = {
-        x <- unname(apply(x, 1, function(q) unname(sort(q)), simplify = FALSE))
+        x <- unname(as.list(as.data.frame(apply(x, 1, function(q) unname(sort(q))))))
```
dajmcdon, line 107:
```r
x <- unname(split(apply(x, 1, function(q) sort(q)), seq(nrow(x)))
```

dajmcdon (changes requested):
Good catch! I think we can simplify further. See if my suggestion below works.

R/make_quantile_reg.R
rachlobay: Yes. I think we could simplify it… though when I replace the code with that line & run the quantile_reg() doc. example, I get this error:

```
Error:
! `q[order(tau)]` produces unsorted quantiles.
Run `rlang::last_trace()` to see where the error occurred.
Called from: signal_abort(cnd, .file)
```
I think that a slight change of order of operations should do the trick... Let’s run through an ex based on the documentation & let me know if it does what you think it should.

Suppose we’re working with this temp (x just before going through that bit of code), which is a 100 x 2 matrix:
```
head(xtemp)
    tau= 0.2   tau= 0.8
1  0.4681302 -1.2231489
2 -1.2828086  0.6921821
3  0.4660981 -0.9436714
4 -1.3417686  0.6614161
5 -1.0832523  0.5037221
6 -1.2014149  0.4241279
```

So, 1st split the df into a list as you said:
```
> head(split(xtemp, seq(nrow(xtemp))))
$`1`
[1]  0.4681302 -1.2231489

$`2`
[1] -1.2828086  0.6921821

$`3`
[1]  0.4660981 -0.9436714

$`4`
[1] -1.3417686  0.6614161

$`5`
[1] -1.0832523  0.5037221

$`6`
[1] -1.2014149  0.4241279
```
Then unname the list:
```
> unname(head(split(xtemp, seq(nrow(xtemp)))))
[[1]]
[1]  0.4681302 -1.2231489

[[2]]
[1] -1.2828086  0.6921821

[[3]]
[1]  0.4660981 -0.9436714

[[4]]
[1] -1.3417686  0.6614161

[[5]]
[1] -1.0832523  0.5037221

[[6]]
[1] -1.2014149  0.4241279
```
And use lapply to sort each pair.
```
> head(lapply(unname(split(xtemp, seq(nrow(xtemp)))), function(q) sort(q)))
[[1]]
[1] -1.2231489  0.4681302

[[2]]
[1] -1.2828086  0.6921821

[[3]]
[1] -0.9436714  0.4660981

[[4]]
[1] -1.3417686  0.6614161

[[5]]
[1] -1.0832523  0.5037221

[[6]]
[1] -1.2014149  0.4241279
```
So all together we get that `x <- lapply(unname(split(x, seq(nrow(x)))), function(q) sort(q))`

dajmcdon (approved):

Issue #172, opened 2023-04-26
Is there a friendlier/consistent way to specify `lags` in `arx_args_list()`?


brookslogan:
@dshemetov notes that the int-vector or list-of-int-vectors-meant-to-be-lined-up-sequentially-with-the-`predictors`-from-`arx_forecaster()` format accepted by `arx_args_list()` could be unintuitive, and differs from how these are specified in `step_epi_lag`.  I don't know if there is a clear alternative, but we might seek something more consistent with `step_epi_lag`, or that enables us to use tidyselections instead of character vectors for `predictors`.

dshemetov:
Vague idea: have the user feed `step_epi_lag` into `arx_args_list` and have `arx_forecaster` apply that step?

dajmcdon:
`step_epi_lag()` applies the same lags to any columns selected by `...`. Because `...` obeys tidyselect syntax. There's no way to know how many columns are in `...`, so this step (like all other steps) must do the same operation to all columns. If you want to have different sets of lags for different sets of predictors, you have to call it multiple times. `arx_forecaster()` internally loops through to achieve the multiple calls. I'm not sure there's another way other than forcing the `arx_args_list()` to require `lags` to be a named list, with names corresponding to the predictors. Seems annoying.

dshemetov:
Ah, if `lags` is a named list like `lags = list(confirmed_incidence_num = c(0, 7, 14), deaths_incidence_num = c(0, 7))`, then that seems good enough for me 👍. For context, I was coming to this function after reading `covidcast::aggregate_signals` where the lags list wasn't named, but was instead numbered, so it required you to keep the signals list in sync with the lags list, which is the kind of mental overhead I dislike having to maintain. But a named list is much better.

dajmcdon:
TODO:

* Add functionality to `arx_args_list()` to check if the lags list is named. If so, validate and order correctly. If not, assume it's the same as the order of the predictors.

PR #218, opened 2023-07-08
172 arx lags list

rachlobay:
Closes Issue #172: Check if the lags list is named. If so, validate and order correctly. If not, assume it's the same as the order of the predictors. Added the relevant code to `arx_lags_validator()` as that seemed to be a good place to have it. Also, added some tests to test-arx_args_list.R.

A couple questions/things to consider…
- I think we should shoot out a warning if we’re assuming that an unnamed list of lags is in the same order as the predictors (I’ve added this).
- I was thinking about partially named lists (ex. name one element but not the others)… If we really want, we can try to handle those (though I think that may become complicated - such as if someone specifies just a few names in a long list). For now, I’ve just treated that as incomplete and, hence, as an unnamed list.

### DIFF
--- a/R/canned-epipred.R
+++ b/R/canned-epipred.R
@@ -22,6 +22,7 @@ validate_forecaster_inputs <- function(epi_data, outcome, predictors) {
 
 arx_lags_validator <- function(predictors, lags) {
   p <- length(predictors)
+
   if (!is.list(lags)) lags <- list(lags)
   l <- length(lags)
   if (l == 1) lags <- rep(lags, p)
@@ -30,6 +31,21 @@ arx_lags_validator <- function(predictors, lags) {
       "You have requested {p} predictor(s) but {l} different lags.",
       i = "Lags must be a vector or a list with length == number of predictors."
     ))
+  }else{
+    if(length(lags) == sum(names(lags) != "", na.rm = TRUE)){
+      if(!all(names(lags) %in% predictors)){
+        cli::cli_abort(
+          "The names for the list of lags do not all correspond to the predictors."
+        )
+      } else{
+        lags <- lags[order(match(names(lags), predictors))]
+      }
+    } else{
+      cli::cli_warn(
+        "The unnamed list of lags has been set to correspond to the order
+          of the predictors.",
+      )
+    }
   }
   lags
 }
--- a/tests/testthat/test-arx_args_list.R
+++ b/tests/testthat/test-arx_args_list.R
@@ -52,5 +52,55 @@ test_that("arx forecaster disambiguates quantiles", {
   expect_error(compare_quantile_args(alist, tlist))
 })
 
+test_that("arx_lags_validator handles named & unnamed lists as expected", {
 
+  # Fully named list of lags in order of predictors
+  pred_vec <- c("death_rate", "case_rate")
+  lags_init_fn <- list(death_rate = c(0, 7, 14), case_rate = c(0, 1, 2, 3, 7, 14))
+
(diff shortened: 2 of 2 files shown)


R/canned-epipred.R
```diff
@@ -30,6 +31,21 @@ arx_lags_validator <- function(predictors, lags) {
       "You have requested {p} predictor(s) but {l} different lags.",
       i = "Lags must be a vector or a list with length == number of predictors."
     ))
+  }else{
```
dajmcdon, line 34: Better linting: `} else {` rather than `}else{`

R/canned-epipred.R
```diff
@@ -30,6 +31,21 @@ arx_lags_validator <- function(predictors, lags) {
...
+    if(length(lags) == sum(names(lags) != "", na.rm = TRUE)){
```
dajmcdon, line 35: Better linting: `if (length(lags)...` rather than `if(length(lags)...`

R/canned-epipred.R
```diff
@@ -30,6 +31,21 @@ arx_lags_validator <- function(predictors, lags) {
...
+      if(!all(names(lags) %in% predictors)){
```
dajmcdon, line 36: Similar linting comment.

Also, let's reverse this `if` `else` pair. First do the one that works else throw the error.

R/canned-epipred.R
```diff
@@ -30,6 +31,21 @@ arx_lags_validator <- function(predictors, lags) {
...
+        )
+      } else{
+        lags <- lags[order(match(names(lags), predictors))]
+      }
+    } else{
+      cli::cli_warn(
+        "The unnamed list of lags has been set to correspond to the order
+          of the predictors.",
+      )
+    }
```
dajmcdon, line 48: Let's remove the warning. I think passing in an unnamed list is fine, and sometimes more convenient. It's also the use case in many examples/vignettes that would become very noisy if this were merged. 

Let's just make sure the documentation says the same.

dajmcdon (changes requested):

R/canned-epipred.R
dajmcdon: And a space before the final `{`

R/canned-epipred.R
```diff
@@ -31,20 +31,15 @@ arx_lags_validator <- function(predictors, lags) {
...
-  }else{
-    if(length(lags) == sum(names(lags) != "", na.rm = TRUE)){
-      if(!all(names(lags) %in% predictors)){
+  } else {
+    if(length(lags) == sum(names(lags) != "", na.rm = TRUE)) {
+      if (all(names(lags) %in% predictors)) {
+        lags <- lags[order(match(names(lags), predictors))]
+      } else {
         cli::cli_abort(
           "The names for the list of lags do not all correspond to the predictors."
```
dajmcdon: Sorry, a few more corrections:
1. On 35: `if(length(lags)...` should be `if (length(lags)...` with a space between `if` and `(`.
2. For the error, let's say "If `lags` is a named list, then all `predictors` must be present". It would be good to also list which ones are missing. An example of how to do this is `validate_meta_match()` in `epi_check_training_set.R`. (Though there, I see that on line 43 it has "`geo_type`" when it should have `{meta}`)
3. Finally, I think the check on 36 is actually not quite right. What we need is all predictors in names(lags) rather than the other way around.

dajmcdon (changes requested):

dajmcdon (approved):

PR #219, opened 2023-07-14
Djm/test data fixes

### LINKED ISSUE #207: get_test_data() recent NA check
It may be checking all variables in the data rather than just those that it needs to access.

dajmcdon:
Closes #207 .

This is a rather involved change.
* Added an argument to `get_test_data()` to better handle NA filling. The goal is to check the forecast date, which can be after then end of the training data, and fill up to that date.
* Added an auxiliary function to add explicit NA rows to the data if the forecast date is after the end of the training data.
* Switch all filling operations to depend on the dates rather than the head/tail of the data frame.
* Properly check whether we can/can't fill the NAs.

This necessitated a number of downstream changes to the canned forecasters. I'm hopeful that there won't be any induced errors in vignettes/tooling book.

Some minor linting.

Added .Rprofile to allow more typical `document()` `load_all()` behaviour (rather than necessitating an `unload("parsnip")` first.


rachlobay:
The changes look ok to me…

A few things to consider about the `get_test_data()` doc. and tests:

- Should we let the user know about `min_required` (ie. what is sufficient training data) in a line in the `get_test_data()` documentation? Perhaps we should give them the heads up about what will give rise to that error.

- Maybe add some simple tests `get_test_data` for the forecast_date, insufficient training data (the major new editions to the function that could cause problems). 

- Also consider updating `get_test_data` doc. to reflect the use of max horizon? Ex. Should the sentence spanning most of lines 21 and 22 say something about max horizon? Because `n_recent` is `NULL` it is set to `min_required + 1` where `min_required` is `max_lags + max_horizon` (around line 74)?

R/get_test_data.R
rachlobay, line 1: See though on this on doc. + tests for this at the end of [here](https://github.com/cmu-delphi/epipredict/pull/219).

rachlobay (approved):

Issue #109, opened 2022-07-20
bug in `layer_add_forecast_date()`


dajmcdon:
The default `forecast_date` (and `target_date`) seems wrong. 

Using `case_death_rate_subset` the fd should default to 2021-12-31 (most recent available) but instead is 2022-01-28. I think this has to do with the fact that the `components$keys$time_value` gets adjusted in the training data to add the `ahead`. We actually need the latest value in the test data (though if there isn't a lag 0, this wouldn't work either).

It may help to have access to the entire workflow at predict time as suggested in #108 .

rachlobay:
Whoops... Posted in #108 106 by accident, so moving those comments here.

Based on the ~July 6 discussion about this on Components conundrum doc, the `time_value` from `components$keys` is currently being used for both forecast_date and test_date defaults. If you want the non-forged test data to be used instead, it still does not look to be currently available in components, as I think you noted. For reference, I will post the ex. from that doc below to show what is available in components for the `layer_add_forecast_date` test (it looks like the molded training data and the forged test data parts are there). If that is the case, components does not look to have what we need. So what is the best way to get the original test data? Has there been something new introduced that would make this a piece of cake because right now, it isn’t clear that `object`, `components`, `the_recipe`, or `the_fit` contains it… And even if we included `the_frosting` in `slather()`, would that contain the test data that is just inputted in `predict()`?  I am not sure about that, so let me know if it does…. Likewise for `the_workflow`. 

We could always revert these back to what I had a long time ago and have the user input the (non-forged) test data into the layer directly as an argument, but I’d have thought that there’d be an easier way for something like this.

rachlobay:
Components conundrum ex. set-up (see test-layer_add_forecast_date.R for this):
```
jhu <- case_death_rate_subset %>%
  dplyr::filter(time_value > "2021-11-01", geo_value %in% c("ak", "ca", "ny"))
r <- epi_recipe(jhu) %>%
  step_epi_lag(death_rate, lag = c(0, 7, 14)) %>%
  step_epi_ahead(death_rate, ahead = 7) %>%
  step_naomit(all_predictors()) %>%
  step_naomit(all_outcomes(), skip = TRUE)
wf <- epi_workflow(r, parsnip::linear_reg()) %>% fit(jhu)
latest <- jhu %>%
  dplyr::filter(time_value >= max(time_value) - 14)
```

What (I think) we want to extract max(time_value):

```
> latest
An `epi_df` object, with metadata:
[...]

PR #220, opened 2023-07-17
Slather access workflow and test data & update `layer_add_target_date()` and `layer_add_forecast_date()` accordingly

### LINKED ISSUE #108: `slather()` needs access to the `frosting` object
Refactor `slather()` throughout so that it takes `the_frosting` as an argument.

We may actually need access to the layers we create.

On further thought, it may be better to make the signature `slather(object, components, the_workflow,...)` so that it has access to everything (rather than `the_recipe` `the_fit` and `the_frosting` separately).

rachlobay:
Fixes #108 and #109 by adding `workflow` and `new_data` to the signature of slather. Then used the `max(new_data$time_value)` as the default forecast date (and that + ahead for the default target date). Finally, updated the tests and documentation to reflect these new defaults.


dajmcdon (changes requested):

R/epi_workflow.R
```diff
@@ -113,14 +155,12 @@ predict.epi_workflow <- function(object, new_data, ...) {
...
   components <- list()
-  the_fit <- workflows::extract_fit_parsnip(object)
-  the_recipe <- workflows::extract_recipe(object)
   components$mold <- workflows::extract_mold(object)
   components$forged <- hardhat::forge(new_data,
                                       blueprint = components$mold$blueprint)
   components$keys <- grab_forged_keys(components$forged,
                                       components$mold, new_data)
-  components <- apply_frosting(object, components, the_fit, the_recipe, ...)
+  components <- apply_frosting(object, components, new_data, ...)
```
dajmcdon, line 163: Don't we still need `the_fit` and `the_recipe`? Am I missing something in the changes?

dajmcdon (changes requested):

R/epi_recipe.R
```diff
@@ -131,6 +131,7 @@ epi_recipe.epi_df <-
       term_info = var_info,
       steps = NULL,
       template = x[1,],
+      mtv = max(x$time_value),
```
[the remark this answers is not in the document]
rachlobay: This should be fixed now.

R/epi_workflow.R
rachlobay: The `epi_workflow` (object) contains the recipe and fit, so it seemed unnecessary to include them separately.

dajmcdon: We might need some more tests then. I would think this would fail for sure.

For the future, when PRs touch lots of files like this one, it's helpful for me if your initial comment goes through the logic of the things you do. That way I can figure out where to look, and if there are simplifications that happen across multiple files, it's easier for me to understand the idea.

dajmcdon (approved):
Everything looks good. A few further changes (these are minor, so I'm approving without them):

1. There's also `layer_unnest()` which needs the `slather()` method correction.
2. `inst/templates/layer.R` contains a helper for creating new layers that needs the same.

rachlobay:
Yes. Good catch for 2! I didn't think to look in inst/templates. So I've updated that, but I don't see 1 (the signature in slather.layer_unnest looks ok, but maybe something inside the function is off? Let me know).

dajmcdon:
Huh. For some reason, I didn't see the `layer_unnest()` as being edited earlier. Must have been an oversight. 

Merge when ready.

Issue #73, opened 2022-06-27
`update_layer()`


dajmcdon:
Create an `update` method for postprocessing. 

See [recipes::update()](https://github.com/tidymodels/recipes/blob/HEAD/R/update.R) for the equivalent preprocessing version.

dajmcdon:
Flagging this up and bumping the priority.

dajmcdon:
We should also be able to update the workflow and the model by going to different `update.method()`s

dajmcdon:
See also #179

Issue #179, opened 2023-05-07
Add a function to output the unfit workflow that would be used by `arx_forecaster()`


dajmcdon:
Alternatively: add a function to output the unfit workflow that would be used by `arx_forecaster()`, have utilities to modify the workflow more easily (`[add_]step*(.after = "<step/phasename>", .before = "<step/phasename>", ....)`?

_Originally posted by @brookslogan in https://github.com/cmu-delphi/epipredict/issues/171#issuecomment-1523959307_

PR #221, opened 2023-07-21
Added `update.layer` and `update_frosting`

rachlobay:
Addresses #73. Added `update.layer()` and `update_frosting()` along with tests for each.


rachlobay:
Actually, hold up... I see that workflows `update_recipe()` and `remove_recipe()` convert the `epi_workflow` to just a `workflow` unfortunately... And I expect that their `update_model()` function will have a similar result... I will have to add the equivalents to work with an `epi_workflow()`. It's probably best to try get this done all on one PR (as they are very similar tasks), so I will let you know/post here after I've done this.

rachlobay:
Update: I’ve finished adding functions (+ tests) for add/update/remove a model and a recipe that work with an `epi_workflow`. Note that I did not add functions to update a recipe step or model specification in a parsnip model because the existing functions for those seem to work ok. 

What I mean by the last sentence can be seen through the following example:
```
library(epiprocess)
library(dplyr)
library(recipes)

jhu <- case_death_rate_subset %>%
  filter(time_value > "2021-08-01") %>%
  dplyr::arrange(geo_value, time_value)

r <- epi_recipe(jhu) %>%
  step_epi_lag(death_rate, lag = c(0, 7, 14)) %>%
  step_epi_ahead(death_rate, ahead = 7) %>%
  step_epi_lag(case_rate, lag = c(0, 7, 14)) %>%
  step_naomit(all_predictors()) %>%
  step_naomit(all_outcomes(), skip = TRUE)

# Update a recipe step using the tidymodels function
# https://github.com/tidymodels/recipes/blob/HEAD/R/update.R
r2 <- r
r2$steps[[1]] <- update(r2$steps[[1]], lag = 4)

# Update model specification for a parsnip model using the parsnip function
# https://parsnip.tidymodels.org/reference/parsnip_update.html
mod1 <- rand_forest(mode = "regression", trees = 10)

update(mod1, trees = 20)
```

R/epi_workflow.R
```diff
@@ -58,6 +58,102 @@ is_epi_workflow <- function(x) {
...
+#' wf
+#'
+#' lm_model <- parsnip::linear_reg()
+#'
+#' wf <- update_model(wf, lm_model)
+#' wf
+#'
+#' wf <- remove_model(wf)
+#' wf
+#' @import workflows
```
dajmcdon, line 111: why import all of `workflows`?

R/layers.R
```diff
@@ -30,6 +30,77 @@ layer <- function(subclass, ..., .prefix = "layer_") {
...
+#'   layer_add_forecast_date(forecast_date = "2022-05-31") %>%
+#'   layer_naomit(.pred)
+#'
+#' wf1 <- wf %>% add_frosting(f)
+#'
+#' p1 <- predict(wf1, latest)
+#' p1
+#'
+#' # Update forecast date
+#' f$layers[[2]] <- update(f$layers[[2]], forecast_date = "2021-06-01")
```
dajmcdon, line 65: I think we should discuss how this is implemented. I don't really like `recipes::update()` because you somehow have to know the order of the layers. I think it should instead look for a class. Perhaps we should have `update_layer()` and `update_epi_recipe()` as more nuanced wrappers around `update()`.

rachlobay: Do you mean to have the user input the name of the layer (like in the `extract_argument` functions)? But then, what if there are two or more layers with the same name (similar to how there may be multiple lag steps)?

dajmcdon: Yes, that's what I mean. And I agree with the potential difficulty. But I feel like the existing version is far too annoying. I think we should brainstorm how to avoid possible bad behaviour while also making it easy for the user.

* There's also the case of trying to update a recipe in a workflow.

* An ideal test bed is something like:
  - take the fitted workflow from `arx_forecaster()`
  - add a recipe step in between other steps
  - change some of the existing steps and layers
  - refit the whole thing and produce predictions

See also #179. I think this is kind of a big project, and I'm not sure how to handle it yet. Maybe good to discuss in a future tooling meeting?

R/epi_workflow.R
rachlobay: You are right. That's not necessary. So I've taken that out.

R/layers.R
rachlobay: This does sound better (and I think something like #179 would be doable), though I also think that it may be good to have something simple to use to be able to update the workflow with in the meantime (rather than the current approach of going down the rabbit hole of $ as in `axe$pre$actions$recipe$recipe$steps[[1]]` - see the discussion in #179)

R/epi_recipe.R
```diff
@@ -273,12 +287,65 @@ is_epi_recipe <- function(x) {
...
+    trained = FALSE
+  )
+}
+
+#' @rdname add_epi_recipe
+#' @export
+update_epi_recipe <- function(
+    x, recipe = NULL, step_num = NULL, ..., blueprint = default_epi_recipe_blueprint()) {
+
+  if(is_null(recipe) && !is_null(step_num)){
```
dajmcdon, line 341: Please add spaces around `if (...) {`

R/epi_recipe.R
```diff
@@ -273,12 +287,65 @@ is_epi_recipe <- function(x) {
...
+    pre = workflows:::new_stage_pre(actions = actions),
+    fit = x$fit,
+    post = x$post,
+    trained = FALSE
+  )
+}
+
+#' @rdname add_epi_recipe
+#' @export
+update_epi_recipe <- function(
```
dajmcdon, line 338: I think we should do this a bit differently. Let's create:

1. `replace_epi_recipe()`. It removes the recipe from the workflow and then adds the new one.
2. `update_epi_recipe()` (maybe even `adjust_epi_recipe()`). This should be an S3 generic with methods `.workflow()` and `.recipe()`. Then args can be `x`, `step_num` (mandatory), `...`, and `blueprint`. Then it can do the update to either kind of object.

R/frosting.R
```diff
@@ -84,8 +109,18 @@ validate_has_postprocessor <- function(x, ..., call = caller_env()) {
   invisible(x)
 }
 
+#' @rdname add_frosting
+#' @export
+update_frosting <- function(x, frosting = NULL, layer_num = NULL, ...) {
 
+  if(is_null(frosting) && !is_null(layer_num)){
```
dajmcdon, line 116: same spacing thing.

R/frosting.R
```diff
@@ -84,8 +109,18 @@ validate_has_postprocessor <- function(x, ..., call = caller_env()) {
...
+    frosting = extract_frosting(x)
```
dajmcdon, line 117: prefer `<-` over `=`

dajmcdon (changes requested):
1. One structural issue which we can discuss.
2. A few minor style issues. 
3. There's a stray file  `R/add_model.Rd` to remove.

R/epi_recipe.R
rachlobay: Done, but instead of `replace_epi_recipe()` used the name `update_epi_recipe()` to maintain consistency with the functions in the recipes package (particularly their `update_recipe()` function from https://github.com/tidymodels/workflows/blob/HEAD/R/pre-action-recipe.R, which I think operates in the same way).
 So, the name used for 2. is `adjust_epi_recipe()` (& correspondingly `adjust_frosting()`).

R/epi_recipe.R
```diff
@@ -276,13 +280,117 @@ is_epi_recipe <- function(x) {
...
+
+  recipe <- workflows::extract_preprocessor(x)
+  recipe$steps[[step_num]] <- update(recipe$steps[[step_num]], ...)
+
+  update_epi_recipe(x, recipe, blueprint = blueprint)
+}
+
+#' @rdname adjust_epi_recipe
+#' @export
+adjust_epi_recipe.epi_recipe <- function(
```
dajmcdon, line 388: I think I've mentioned this, but I'd like to see this allow for `step_name` as well. You can access these with something like:
`map_chr(x$steps, ~ attr(.x, "class")[1])`.

Then if it's a unique match, use it. Otherwise, return an error that gives the numbers that would have been available.

dajmcdon: Also worth mentioning that one can use `tidy(recipe)` to see some summary information about the recipe. We should make a similar method for the frosting. And provide some documentation (short Vignette?) that illustrates all of these as well as updating.

R/epi_workflow.R
```diff
@@ -58,6 +58,103 @@ is_epi_workflow <- function(x) {
...
+
+#' @rdname add_model
+#' @export
+remove_model <- function(x)
+  UseMethod('remove_model')
+
+#' @rdname add_model
+#' @export
+update_model <- function(x, spec, ..., formula = NULL)
+  UseMethod('update_model')
```
dajmcdon, line 124: Please run `styler::style_pkg()`. It will handle the missing `{` and `}` for these.

R/epi_workflow.R
```diff
@@ -58,6 +58,103 @@ is_epi_workflow <- function(x) {
...
+
+#' @rdname add_model
+#' @export
+add_model.epi_workflow <- function(x, spec, ..., formula = NULL) {
+  workflows::add_model(x, spec, formula = formula)
```
dajmcdon, line 129: needs to pass along `...`

R/frosting.R
```diff
@@ -27,10 +27,17 @@
 #' p1 <- predict(wf1, latest)
 #' p1
 #'
-#' # Remove frosting from the workflow and predict
-#' wf2 <- wf1 %>% remove_frosting()
+#' # Update frosting in a workflow and predict
```
dajmcdon, line 30: one of these examples should do something that results in different predictions.

R/frosting.R
```diff
@@ -87,8 +93,83 @@ validate_has_postprocessor <- function(x, ..., call = caller_env()) {
...
+#' f2 = f1 %>% adjust_frosting(layer_num = 2, upper = 5)
+#' extract_frosting(wf2)
+#'
+adjust_frosting <- function(x, layer_num, ...) {
+  UseMethod("adjust_frosting")
+}
+
+#' @rdname adjust_frosting
+#' @export
+adjust_frosting.epi_workflow <- function(
```
dajmcdon, line 156: Same idea with adjusting based on names (as well as numbers) as above.

Probably the best way is to replace `layer_num` with `which_layer`. If it's numeric, then great. If it's character, then check for uniqueness.

vignettes/preprocessing-and-models.Rmd
```diff
@@ -245,6 +245,7 @@ State-wise population data from the 2019 U.S. Census is included in this package
 and will be used in `layer_population_scaling()`.
 
 ```{r}
+
```
dajmcdon, line 248: I don't think we need this empty line...

dajmcdon (changes requested):
I added a number of comments. The ideas are good but a few major changes:

* Add a `tidy()` method for `frosting` objects.
* Allow the adjust methods to handle names as well as numbers.
* Create a short vignette that illustrates how these can be used. (including the `tidy()` methods for examining the recipe/frosting).
* Run styler so that the code formatting matches the style on main.

dajmcdon (approved):

PR #223, opened 2023-08-03
docs: add equivalent pipeline to arx_forecaster docs

dsweber2:
this is mostly(?) in the getting started inner workings section, but it would be good to have consolidated in the `arx_forecaster` reference docs.

I'm leaving this as a draft in case there are other doc sections I have specific suggestions for


dsweber2:
digging a little further, it is probably worth having another prototype for the other workflow generated (for quantgen), and making it clear that they're *not* the same

PR #226, opened 2023-08-10
Formatting

dsweber2:
This just applies `style_pkg()` to the whole package. There's also a github action to automatically format any push; I'm less certain that will 100% work, because testing github actions is kind of a mess.


dsweber2:
The last error is just because the styler doesn't have permissions to actually make a commit. The previous one was because of the same error as my other pull request, but this is now strictly based on that one

dajmcdon:
Just to be clear, this PR seems to do 3 things:
1. Add `.git-blame-ignore-revs`. There's a commit hash in that file. How exactly does this work?
2. Add a Github action that should automatically style all future PRs.
3. Apply the action to the current `main`.

Is that right?

dsweber2:
Yes, that's a good summary.
it's very similar to .gitignore. if the flag --ignore-revs is passed to git blame, or other git history commands, any change from that commit is ignored. This is to make it so that not everything is attributed to who or whatever is doing the styling, and history diffs can ignore it (much like they would other whitespace changes).

dajmcdon:
@dsweber2 can you handle the conflicts above, or should I?

dsweber2:
Got it; best handled via local rebase; I also added the revised commit to `.git-ignore-revs`.

The tests are broken on main btw. I think I reproduced exactly the same errors, but its hard to tell.

dajmcdon:
If you merge `main` into this, then checks should pass. I'll merge this once it's ready.

dsweber2:
alright, tests pass! And the new code on main is formatted

dajmcdon (approved):
OK. Merge when ready.

PR #230, opened 2023-09-06
In printing of an `epi_workflow`, number steps and layers

rachlobay:
Made it so that steps (and layers) are numbered rather than by bullets in the printing of a `epi_workflow` using `cli:::cli_ol` as discussed.


R/frosting.R
```diff
@@ -297,7 +285,7 @@ print_frosting <- function(x, ...) {
   layer_names <- map_chr(layers, pull_layer_name)
 
   if (n_layers <= 10L) {
-    cli::cat_bullet(layer_names)
+    cli:::cli_ol(layer_names)
```
dajmcdon, line 288: Why `:::` instead of `::`?

R/frosting.R
```diff
@@ -269,18 +269,6 @@ apply_frosting.epi_workflow <-
   }
 
 #' @export
```
dajmcdon, line 271: We still want a print method for frosting (it can just call `print_frosting()`). That way if you have a frosting object, it won't just print the unformatted list.

R/epi_recipe.R
```diff
@@ -397,3 +397,77 @@ kill_levels <- function(x, keys) {
   for (i in which(names(x) %in% keys)) x[[i]] <- list(values = NA, ordered = NA)
   x
 }
+
+#' @export
```
dajmcdon, line 401: This probably shouldn't be exported. But we also want `print.epi_recipe()` which _should_ be exported. Currently, because an `epi_recipe` is a subclass of `recipe`, calling `print(epi_recipe)` will dispatch to `print.recipe()`. If we want numbered, then we need to make a method.

dajmcdon (changes requested):

R/frosting.R
rachlobay: Switched to `::`

R/epi_recipe.R
```diff
@@ -410,3 +410,132 @@ kill_levels <- function(x, keys) {
...
+print.epi_recipe <- function(x, form_width = 30, ...) {
+  cli::cli_div(theme = list(.pkg = list("vec-trunc" = Inf, "vec-last" = ", ")))
+
+  cli::cli_h1("Recipe")
```
dajmcdon, line 418: Epi Recipe

dajmcdon:
@rachlobay Is this ready to go now?

rachlobay:
Yes, I've made the two changes discussed on Tuesday.

dajmcdon (approved):

PR #236, opened 2023-09-13
axe the outdated musings directory, it still lives in various places,…

dajmcdon:
… including the `stored-musings` branch


dajmcdon:
@dsweber2 rather than forcing a merge, can you just ensure that I'm not missing something? I think I removed only the offending directory.

dsweber2 (approved):
I only see removals in musings, so lgtm? We'll have to pull this into formatting if we merge it before that branch, but that's not exactly hard.

TIL you can put `.gitignore` is subdirectories.

dajmcdon:
Wait for #226 .

Issue #237, opened 2023-09-13
make layer for hub-submission format


dsweber2:
Was a summer OKR, also seems like something CDC would like.

brookslogan:
Progress in #245.  This does not handle categorical rate-trend forecasts, requires use of covidcast `geo_value` naming scheme, and #249.

PR #241, opened 2023-09-25
240 quantile pivot

### LINKED ISSUE #240: Refactor quantile pivoting
`pivot_quantiles()` only does "wide".

- [x] rename current version to indicate "wide".
- [x] add another version to do "long"

dajmcdon:
Closes #240 

@dsweber2 This isn't a "layer" but you should be able to call it directly on the predictions and get out something much closer to the submission format.

* There was already the `_wider()` version. But naming is better.
* The `_longer()` version should be what you want.

This function is a necessary precursor to #237


dsweber2 (approved):
awesome, thanks for this! If it isn't too gnarly, is it possible to change the column names from `q` and `tau` to something user defined/default to quantile/value? If not, it's easy enough to pass this through `rename` after

dajmcdon:
On the `q`/`tau` issue, this isn't the only place in the package we have this problem. See also #147 (now 1 year old). We need to standardize these throughout. I'll put it on the Tooling agenda.

dajmcdon:
This is a breaking change. Needs a version bump and/or a shim for the dperecated `pivot_quantiles()`.

R/pivot_quantiles.R
```diff
@@ -0,0 +1,160 @@
...
+        i = "Check failed for variables(s) {.var {nms}}."
+      )
+    )
+  }
+  if (length(cols) > 1L) {
+    for (col in cols) {
+      .data <- .data %>%
+        tidyr::unnest(tidyselect::all_of(col)) %>%
+        tidyr::pivot_wider(
+          names_from = "tau", values_from = "q",
```
nmdefries, line 133: thought: These names were updated in https://github.com/cmu-delphi/epipredict/pull/243 but since the functions got moved around here vs that branch, merging them will probably get hairy. The https://github.com/cmu-delphi/epipredict/pull/243 changes might not percolate to this new file, so reminder to check.

R/pivot_quantiles.R
```diff
@@ -0,0 +1,160 @@
...
+#'
+#' @return An object of the same class as `.data`.
+#' @export
+#'
+#' @examples
+#' d1 <- c(dist_quantiles(1:3, 1:3 / 4), dist_quantiles(2:4, 1:3 / 4))
+#' d2 <- c(dist_quantiles(2:4, 2:4 / 5), dist_quantiles(3:5, 2:4 / 5))
+#' tib <- tibble::tibble(g = c("a", "b"), d1 = d1, d2 = d2)
+#'
+#' pivot_quantiles_longer(tib, "d1")
```
nmdefries, line 52: note: Want to verify that this is the expected behavior:

Screenshot from 2023-10-04 17-15-56

I see that the `q` and `tau` columns are inserted at the spot where the old column used to be, instead of always ending up at the end, for example.

nmdefries (approved):

NEWS.md
```diff
@@ -7,7 +7,7 @@
 * canned forecasters get a class
 * fixed quantile bug in `flatline_forecaster()`
 * add functionality to output the unfit workflow from the canned forecasters
-* add `pivot_quantiles()` for easier plotting
+* add `pivot_quantiles_wider()` for easier plotting
```
nmdefries, line 10: question: should this also note the `*_longer` version?

R/pivot_quantiles.R
```diff
@@ -0,0 +1,160 @@
...
+#'   expressions separated by commas. Variable names can be used as if they
+#'   were positions in the data frame, so expressions like `x:y` can
+#'   be used to select a range of variables.
+#' @param .ignore_length_check If multiple columns are selected, as long as
+#'   each row has contains the same number of quantiles, the result will be
+#'   reasonable. But if, for example, `var1[1]` has 5 quantiles while `var2[1]`
+#'   has 7, then the only option would be to recycle everything, creating a
+#'   _very_ long result. By default, this would throw an error. But if this is
+#'   really the goal, then the error can be bypassed by setting this argument
+#'   to `TRUE`. The first selected column will vary fastest.
```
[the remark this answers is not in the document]
dajmcdon: For recycling, if `c(1,2)` and `c(1,2,3)` get recycled to matching sizes, then we'd get `c(1,2,1,2,1,2)` and `c(1,1,2,2,3,3)`, so the first "varies fastest".

R/pivot_quantiles.R
dajmcdon: Definitely hairy, unfortunately.

R/pivot_quantiles.R
dajmcdon: Yes, this is expected (matches `unnest()`).

PR #243, opened 2023-10-01
242 quantile renaming

### LINKED ISSUE #242: Rename quantiles / values for consistency
Creating an explicit issue for the second point in #147.

The decision is `q` -> `values`, `tau` -> `quantile_values` throughout. `probs` also becomes `quantile_values`. Any instance of `levels` was likely a mistake and should be `quantile_values`.

This will be breaking and require a version change.

dajmcdon:
closes #242


dshemetov:
Our meeting notes say we were going to use the term `quantile_level`. Want to make sure that we're consciously switching to `quantile_value` instead. I think `level` better evokes the fixed values of a categorical variable.

dsweber2:
let's see, the variables that were renamed were:
- `levels`
  There's a couple of places with levels still present as a variable:
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/epi_juice.R#L29
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/epi_recipe.R#L125
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/bake.epi_recipe.R#L89
  looking at the set, I think this might just be a clobbered variable name.
- `tau`
  several of these lurking in  `make_smooth_quantile_reg` and `make_quantile_reg` 
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/make_smooth_quantile_reg.R#L127
https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/make_quantile_reg.R#L107
  I'm guessing functions like `tau_duplication` and maybe all the `taus` in dist quantiles are left intentionally?
- `q`
  both `q` and `p` are pretty hard to search for so I may have missed a couple. this function and the one below it both use `q` internally
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/dist_quantiles.R#L33
  another pair down a ways in the same file
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/dist_quantiles.R#L129
  this bit of docs
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/dist_quantiles.R#L155
  actually shows up quite a bit in this file; I suppose given that it's all specifically about quantiles it's probably a more reasonable use.
  This seems like it might actually need to change (also on 127 below that)
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/layer_residual_quantiles.R#L119
  
- `p`
  also quite difficult to search for. I don't think I found any that are probabilities; there's some predictors in examples named `p`, but that's not caught in this refactor. There's this anonymous function, which seems fine
  https://github.com/cmu-delphi/epipredict/blob/07600a28339886b83266772b0d8c6a7e3a3caa24/R/make_smooth_quantile_reg.R#L174
  So dist_quantiles has a number that I think might need to change:
  https://github.com/cmu-delphi/epipredict/blob/242-quantile-renaming/R/dist_quantiles.R#L48
  this one *definitely* needs to change
  https://github.com/cmu-delphi/epipredict/blob/242-quantile-renaming/R/dist_quantiles.R#L123
  lines 334, 338, 342, 345, 358, 363, 364, 365, 376 all seem like reasonable local variables (they're short enough functions that 1 letter variables is probably not the worst thing in the world).
  I'm assuming we don't have doctests or something like that? This one definitely needs to change
  https://github.com/cmu-delphi/epipredict/blob/015b0ea881c3ce99d047fe58ea064a531d5865d8/vignettes/epipredict.Rmd#L168

The search I ended up using for the single letters was `[^a-zA-Z_\-0-9]p[^a-zA-Z_\-0-9]`, which I _think_ gets rid of all the places that `p` can be part of another word or variable?

dajmcdon:
Hmmm, nuts. That's not what I listed in #242. But my memory is hazy. Does `quantile_level` sound right to the rest of you?

dajmcdon:
I think some of these actually have to stay.
* `levels` this is referring to levels of a factor.
* `tau` these are the underlying arguments of functions from other packages. This syntax actually maps our new names for these things to those arguments. So these stay.
* `q` I think these are all local. I'm happy to try to change them, but I don't think it will matter.
* `p` These are a bit stupid. In multiple cases, I used partial argument matching. The generic `quantile()` function takes `probs` as an argument. So I left `probs` as an arg to `extrapolate_quantiles()` which is basically trying to take quantiles of a `dist_quantiles` object. But you can match the argument as `quantile(x, p = z)` which internally becomes `quantile(x, probs = z)`. This is probably not optimal coding practice. I'll try to fix these.

dsweber2:
- good to know, I just wanted to confirm for both the `levels` and the `tau`
- it might be good for consistency's sake to rename the `q`'s, but I don't have strong feelings
- oh right, partial matching. Thanks for ditching those, they keep causing spurious warnings. Definitely an R footgun imo

dajmcdon:
Ok. I think I fixed all the `quantile_values` -> `quantile_levels`. 

I was wrong about `p` being a partial match. In the `{distributional}` package that we Import for this purpose, it actually defines the `quantile` method as `quantile(x, p, ...)` rather than `quantile(x, probs, ...)`. So we're stuck with that one.

dsweber2 (approved):

dshemetov:
Ask `distributional` maintainer to switch the naming?

nmdefries (commented):
Found a couple vars in `dist_quantiles.R` that may need to be updated, I couldn't tell if David's previous comments addressed these.

PR #246, opened 2023-10-04
CI: add CI button

dsweber2:
Not sure why adding a commit on #243 didn't trigger the action; this is adding a button to do it manually when you want


dshemetov (approved):

dsweber2:
This is not a bad idea, but it isn't necessary to fix the problem we were having

PR #257, opened 2023-10-25
Update _pkgdown.yml

rachlobay:
To include add_model, adjust_epi_recipe, tidy.frosting (else fail pkgdown check).


dajmcdon:
See #258

PR #271, opened 2023-11-29
simplify grab_residuals

### LINKED ISSUE #270: Checking whether grab_residuals method exists eating up unreasonable amounts of time
@dajmcdon as requested. Raised in relation to discussing https://github.com/cmu-delphi/exploration-tooling/issues/60. It appears that the `utils::methods` function [here](https://github.com/cmu-delphi/epipredict/blob/378577a213aa59043dea73e8fdea725432f32f16/R/layer_residual_quantiles.R#L138) is taking 9s out of 90s for running epipredict inside `epix_slide`, which was surprisingly long.

dajmcdon:
closes #270 .

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: epipredict
 Title: Basic epidemiology forecasting methods
-Version: 0.0.6
+Version: 0.0.7
 Authors@R: c(
     person("Daniel", "McDonald", , "daniel@stat.ubc.ca", role = c("aut", "cre")),
     person("Ryan", "Tibshirani", , "ryantibs@cmu.edu", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # epipredict (development)
 
+# epipredict 0.0.7
+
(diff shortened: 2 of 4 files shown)


dsweber2 (approved):
oh, it returns `NULL` if there aren't residuals? that does make this way easier. I'm assuming there's at least one non-regression model calling grab_residuals somewhere in the tests?

dajmcdon:
I kind of doubt that there is a test where a classifier calls this function. Are you thinking that there needs to be?

dshemetov:
Can we bump minor/patch version and add a note to CHANGELOG?

dajmcdon:
@dshemetov Sure, though I'm unfamiliar. What would you like to bump to? 0.0.7 or 0.0.6001? Something else?

dshemetov:
0.0.7 sounds good to me.

dshemetov:
[This](https://usethis.r-lib.org/reference/use_version.html) might do it for you (haven't tried).

dajmcdon:
residuals are just the difference between the training response and the point prediction (at least the kind we're using here), so would be mathematically defined for any regression engine (including quantile regression). This should fail if the engine is a classifier though.

PR #273, opened 2023-11-29
Add documentation for `arx_forecaster` and `sliding`

XuedaShen:
Added `arx_forecaster` only produces predictions for the largest time_value + ahead; and trains one model over all geo_value.

Specified how `before` and `n_traning`
controls number of training samples in `sliding`

### DIFF
--- a/R/arx_forecaster.R
+++ b/R/arx_forecaster.R
@@ -2,7 +2,10 @@
 #'
 #' This is an autoregressive forecasting model for
 #' [epiprocess::epi_df] data. It does "direct" forecasting, meaning
-#' that it estimates a model for a particular target horizon.
+#' that it estimates a model for a particular target horizon. The
+#' forecaster trains one model using data over all available
+#' geo_value, and produces forecasts for `target_date = max(epi_df$time_value)
+#' + arx_arg_list()$ahead`.  
 #'
 #'
 #' @param epi_data An `epi_df` object
--- a/vignettes/articles/sliding.Rmd
+++ b/vignettes/articles/sliding.Rmd
@@ -309,6 +309,11 @@ k_week_version_aware <- function(ahead = 7, version_aware = TRUE) {
         args_list = arx_args_list(ahead = ahead)
       ) %>%
         extract2("predictions"),
+      # Both `before` in `epi_slide` and `n_training` in `arx_forecaster()`
+      # Controls number of traning samples forecaster can see
+      # `before` first controls the number of samples at a particular step 
+      # of `epi_slide`, then `n_training` further restricts max training sample
+      # in that partiuclar step `arx_forecaster()` could use.
       before = 120 - 1,
       ref_time_values = fc_time_values,
       new_col_name = "fc"


R/arx_forecaster.R
```diff
@@ -2,7 +2,10 @@
 #'
 #' This is an autoregressive forecasting model for
 #' [epiprocess::epi_df] data. It does "direct" forecasting, meaning
-#' that it estimates a model for a particular target horizon.
+#' that it estimates a model for a particular target horizon. The
+#' forecaster trains one model using data over all available
+#' geo_value, and produces forecasts for `target_date = max(epi_df$time_value)
```
dajmcdon, line 7: keys (the `geo_value` and any additional)

R/arx_forecaster.R
```diff
@@ -2,7 +2,10 @@
...
+#' + arx_arg_list()$ahead`.
```
dajmcdon, line 8: This is almost the default, but not the only option. See [here](https://github.com/XuedaShen/epipredict/blob/26afaf58d7be5a29a55cdac0f080e39aa9347802/R/arx_forecaster.R#L134). And `ahead` is an argument that can be set inside `arx_arg_list()` as is `forecast_date` or `target_date`. Maybe better to point to that documentation, or an example?

vignettes/articles/sliding.Rmd
```diff
@@ -309,6 +309,11 @@ k_week_version_aware <- function(ahead = 7, version_aware = TRUE) {
         args_list = arx_args_list(ahead = ahead)
       ) %>%
         extract2("predictions"),
+      # Both `before` in `epi_slide` and `n_training` in `arx_forecaster()`
+      # Controls number of traning samples forecaster can see
+      # `before` first controls the number of samples at a particular step
```
dajmcdon, line 314: `n_training` isn't an argument to `arx_forecaster()` but to `arx_args_list()`. 

I think it's OK to document `before` here like `# controls how many days in the `epi_df` are used (see ...)`.

But the rest is really details that should go in the text of the Vignette with additional details.

dajmcdon (changes requested):

vignettes/articles/sliding.Rmd
XuedaShen: Yes, thanks for pointing out `n_training`, I wasn't most precise there. 

Let's drop the additional text on `before` and `n_training` to maybe an example with more details.

XuedaShen: I am actually not too sure about the 

> n_training
> Integer. An upper limit for the number of rows per key that are used for training (in the time unit of the epi_df).

What is the key referring to here? Is it the key in `quantile_by_key`?

R/arx_forecaster.R
XuedaShen: I see. Maybe we can something like by default predicts for `target_date = max(epi_df$time_value) + arx_arg_list()$ahead`.
And add an example that changes the `target_date`?

vignettes/articles/sliding.Rmd
dajmcdon: `epi_df`'s can contain `additional_keys` besides just `geo_value` (say age bins).

PR #274, opened 2023-12-02
Vignette on the add_*/remove_*/update_*, adjust_* and tidy functions

rachlobay:
Here's a vignette that goes over how to use the add/remove/update, adjust and tidy functions that were recently added to the package.

### DIFF
--- a/vignettes/update.Rmd
+++ b/vignettes/update.Rmd
@@ -287,8 +287,8 @@ Note that when we adjust the `r2` object directly, we are not adjusting the
 recipe in the `epi_workflow`. That is, if we modify a step in `r2`, the change
 will not automatically transfer over to `wf`. We would need to modify the recipe
 in `wf` directly (`adjust_epi_recipe()` on `wf`) or update the recipe in `wf`
-with a new `epi_recipe` that has undergone the adjustment (using
-`update_epi_recipe()`):
+with a new `epi_recipe` that has undergone the adjustment 
+(using `update_epi_recipe()`):
 
 ```{r}
 r2 <- adjust_epi_recipe(r2, which_step = 2, lag = 0:21)
--- a/vignettes/update.Rmd
+++ b/vignettes/update.Rmd
@@ -21,25 +21,52 @@ library(epipredict)
 library(recipes)
 ```
 
-In this vignette, we will state the main goal of the add/update/remove and adjust functions and describe what part of the processing each function is intended for. We will then demonstrate how to use the sets of add/update/remove functions, followed by the adjust functions, and end with a brief discussion on the tidy methods for recipe and frosting objects. 
+In this vignette, we will state the main goal of the add/update/remove and
+adjust functions and describe what part of the processing each function is
+intended for. We will then demonstrate how to use the sets of add/update/remove
+functions, followed by the adjust functions, and end with a brief discussion on
+the tidy methods for recipe and frosting objects.
 
 ## Main goal of the add/update/remove and adjust functions
 
-The primary goal of the update and adjust functions is to allow the user to modify a `step`, `layer`, `epi_recipe`, `frosting`, or a part of an `epi_workflow` so that they do not have to create a new object each time they wish to make a change to the pre-processing, fitting, or post-processing.
+The primary goal of the update and adjust functions is to allow the user to
+modify a `step`, `layer`, `epi_recipe`, `frosting`, or a part of an
+`epi_workflow` so that they do not have to create a new object each time they
+wish to make a change to the pre-processing, fitting, or post-processing.
 
-In the context of pre-processing, the goal of the update functions is to add/remove/update an `epi_recipe` or a step in it. For this, we have `add_epi_recipe`, `update_epi_recipe`, and `remove_epi_recipe` to add/update/remove an entire `epi_recipe` in an `epi_workflow` as well as `adjust_epi_recipe` to adjust a particular step in an `epi_recipe` or `epi_workflow` by the step number or name. For a model, one may `add_model`, `update_model`, or `remove_model` in an `epi_workflow`. For post-processing, where the goal is to update a frosting object or a layer in it, we have `add_frosting`, `remove_frosting`, and `update_frosting` to add/update/remove an entire `frosting` object in an `epi_workflow` as well as `adjust_frosting` to adjust a particular layer in a `frosting` or `epi_workflow` by its number or name. A summary of the function uses by processing step is shown by the following table:
+In the context of pre-processing, the goal of the update functions is to
+add/remove/update an `epi_recipe` or a step in it. For this, we have
+`add_epi_recipe()`, `update_epi_recipe()`, and `remove_epi_recipe()` to
+add/update/remove an entire `epi_recipe` in an `epi_workflow` as well as
+`adjust_epi_recipe()` to adjust a particular step in an `epi_recipe` or
+`epi_workflow` by the step number or name. For a model, one may `add_model()`,
+`update_model()`, or `remove_model()` in an `epi_workflow`. For post-processing,
+where the goal is to update a frosting object or a layer in it, we have
+`add_frosting()`, `remove_frosting()`, and `update_frosting()` to
+add/update/remove an entire `frosting` object in an `epi_workflow` as well as
+`adjust_frosting()` to adjust a particular layer in a `frosting` or
+`epi_workflow` by its number or name. A summary of the function uses by
+processing step is shown by the following table:
 
-|                            | Add/update/remove functions                                | adjust functions    |
+| | Add/update/remove functions | adjust functions |
 |----------------------------|------------------------------------------------------------|---------------------|
-| Pre-processing  | `add_epi_recipe`, `update_epi_recipe`, `remove_epi_recipe` | `adjust_epi_recipe` |
-| Model specification        | `add_model`, `update_model` `remove_model`                 |                     |
-| Post-processing | `add_frosting`, `remove_frosting`, `update_frosting`       | `adjust_frosting`   |
+| Pre-processing | `add_epi_recipe()`, `update_epi_recipe()`,
+`remove_epi_recipe()` | `adjust_epi_recipe()` | | Model specification |
+`add_model()`, `update_model()` `remove_model()` | | | Post-processing |
+`add_frosting()`, `remove_frosting()`, `update_frosting()` | `adjust_frosting()`
+|
 
-Since adding/removing/updating frosting as well as adjusting a layer in a `frosting` object proceeds in the same way as performing those tasks on an `epi_recipe`, we will focus on implementing those for an `epi_recipe` in this vignette and only briefly go through some examples for a `frosting` object.
+Since adding/removing/updating frosting as well as adjusting a layer in a
+`frosting` object proceeds in the same way as performing those tasks on an
+`epi_recipe`, we will focus on implementing those for an `epi_recipe` in this
+vignette and only briefly go through some examples for a `frosting` object.
 
 ## Add/update/remove an `epi_recipe` in an `epi_workflow`
 
-We start with the built-in `case_death_rate_subset` dataset that contains JHU daily COVID-19 cases and deaths by state and take a subset of it from Nov. 1, 2021 to Dec. 31, 2021 for the four states of Alaska, California, New York, and South Carolina.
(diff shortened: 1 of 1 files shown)


vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+  out.width = "100%"
+)
+```
+
+```{r setup, message=FALSE}
+library(epipredict)
+library(recipes)
+```
+
+In this vignette, we will state the main goal of the add/update/remove and adjust functions and describe what part of the processing each function is intended for. We will then demonstrate how to use the sets of add/update/remove functions, followed by the adjust functions, and end with a brief discussion on the tidy methods for recipe and frosting objects.
```
dajmcdon, line 24: Can you add carriage returns so that this line breaks at 80 characters? Makes reviewing and editing much easier. Same for much of the below.

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+```{r setup, message=FALSE}
+library(epipredict)
+library(recipes)
+```
+
+In this vignette, we will state the main goal of the add/update/remove and adjust functions and describe what part of the processing each function is intended for. We will then demonstrate how to use the sets of add/update/remove functions, followed by the adjust functions, and end with a brief discussion on the tidy methods for recipe and frosting objects. 
+
+## Main goal of the add/update/remove and adjust functions
+
+The primary goal of the update and adjust functions is to allow the user to modify a `step`, `layer`, `epi_recipe`, `frosting`, or a part of an `epi_workflow` so that they do not have to create a new object each time they wish to make a change to the pre-processing, fitting, or post-processing.
```
dajmcdon, line 28: for functions, it's better to end them with open+close parenthesis. I think here, these are all "objects", so it doesn't need that, but maybe clearer then to add "objects" in the sentence.

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+
+|                            | Add/update/remove functions                                | adjust functions    |
+|----------------------------|------------------------------------------------------------|---------------------|
+| Pre-processing  | `add_epi_recipe`, `update_epi_recipe`, `remove_epi_recipe` | `adjust_epi_recipe` |
```
dajmcdon, line 34: more functions.

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+To see this, let's look at what happens if we remove our current `epi_recipe` using `workflows::remove_recipe` and then inspect the class of `wf`:
+
+```{r}
+wf %>% class() # class before
+workflows::remove_recipe(wf) %>% class() # class after removing recipe using workflows function
+```
+
+We can observe that `wf` is no longer an `epi_workflow` and a `workflow`. It has been demoted to only a `workflow`. While all `epi_workflow`s are `workflow`s, not all `workflow`s are `epi_workflow`s, meaning that there may be compatibility issues and limitations to the tools that may be used from the `epipredict` package on a plain `workflow` object. 
+
+Now, while we checked what happens to the above `epi_recipe` if we remove it, note that we did not actually store that change to `wf` (using the assignment operator `<-`). Hence, our `epi_workflow` remains unchanged.
```
dajmcdon, line 105:
```suggestion
Now, while we checked what happens to the above `epi_recipe` if we remove it, note that we did not actually store that change to `wf`. 
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+
+In the context of pre-processing, the goal of the update functions is to add/remove/update an `epi_recipe` or a step in it. For this, we have `add_epi_recipe`, `update_epi_recipe`, and `remove_epi_recipe` to add/update/remove an entire `epi_recipe` in an `epi_workflow` as well as `adjust_epi_recipe` to adjust a particular step in an `epi_recipe` or `epi_workflow` by the step number or name. For a model, one may `add_model`, `update_model`, or `remove_model` in an `epi_workflow`. For post-processing, where the goal is to update a frosting object or a layer in it, we have `add_frosting`, `remove_frosting`, and `update_frosting` to add/update/remove an entire `frosting` object in an `epi_workflow` as well as `adjust_frosting` to adjust a particular layer in a `frosting` or `epi_workflow` by its number or name. A summary of the function uses by processing step is shown by the following table:
```
dajmcdon, line 30: For example:

```suggestion
In the context of pre-processing, the goal of the update functions is to add/remove/update an 
`epi_recipe` or a step in it. For this, we have `add_epi_recipe()`, `update_epi_recipe()`, and `remove_epi_recipe()` to alter an entire `epi_recipe` in an `epi_workflow` as well as 
`adjust_epi_recipe()` to alter the arguments to a particular step in an `epi_recipe` or `epi_workflow` by the step number or name. 

For a model, one may `add_model()`, `update_model()`, or `remove_model()` in an `epi_workflow`. 

For post-processing, where the goal is to update a frosting object or a layer in it, we have `add_frosting()`, `remove_frosting()`, and `update_frosting()` to alter the entire `frosting` object in an `epi_workflow` as 
well as `adjust_frosting()` to alter the arguments to a  particular layer in a `frosting` or `epi_workflow` by 
its number or name. 

A summary of the function uses by processing step is shown by the following table:
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+  echo = TRUE,
+  collapse = TRUE,
+  comment = "#>",
+  out.width = "100%"
+)
+```
+
+```{r setup, message=FALSE}
+library(epipredict)
+library(recipes)
```
dajmcdon, line 21:
```suggestion
library(recipes)
library(dplyr)
library(workflows)
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+
+Since adding/removing/updating frosting as well as adjusting a layer in a `frosting` object proceeds in the same way as performing those tasks on an `epi_recipe`, we will focus on implementing those for an `epi_recipe` in this vignette and only briefly go through some examples for a `frosting` object.
+
+## Add/update/remove an `epi_recipe` in an `epi_workflow`
+
+We start with the built-in `case_death_rate_subset` dataset that contains JHU daily COVID-19 cases and deaths by state and take a subset of it from Nov. 1, 2021 to Dec. 31, 2021 for the four states of Alaska, California, New York, and South Carolina.
+
+```{r}
+jhu <- case_death_rate_subset %>%
+  dplyr::filter(time_value >= as.Date("2021-11-01"), geo_value %in% c("ak", "ca", "ny", "sc"))
```
dajmcdon, line 46:
```suggestion
  filter(time_value >= as.Date("2021-11-01"), geo_value %in% c("ak", "ca", "ny", "sc"))
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+  step_epi_naomit()
+
+wf <- update_epi_recipe(wf, r2)
+wf
+```
+You can see that the output of `wf` depicts the sequence of steps in `r2` instead of `r`, which indicates that the update was successful.
+
+A longer approach to achieve the same end is to use `remove_epi_recipe` to remove the old recipe and then `add_epi_recipe` to add the new one. Under the hood, the `update_epi_recipe` function operates in this way. 
+
+The `add_epi_recipe` and `remove_epi_recipe` functions offload to the `workflows` versions of the functions (`workflows::add_recipe()` and `workflows::remove_recipe()`) as much as possible. The main reason for using the `epipredict` version is so that we ensure that we do not lose the `epi_workflow` class.
```
dajmcdon, line 94: I prefer package names to go in curly braces. (Good use of paren on the function names here.)

```suggestion
The `add_epi_recipe()` and `remove_epi_recipe()` functions offload to the `{workflows}` versions of the 
functions  as much as possible. The main reason for using the `{epipredict}` version is so that we ensure 
that we retain the `epi_workflow` class. 
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+
+A longer approach to achieve the same end is to use `remove_epi_recipe` to remove the old recipe and then `add_epi_recipe` to add the new one. Under the hood, the `update_epi_recipe` function operates in this way. 
+
+The `add_epi_recipe` and `remove_epi_recipe` functions offload to the `workflows` versions of the functions (`workflows::add_recipe()` and `workflows::remove_recipe()`) as much as possible. The main reason for using the `epipredict` version is so that we ensure that we do not lose the `epi_workflow` class. 
+
+To see this, let's look at what happens if we remove our current `epi_recipe` using `workflows::remove_recipe` and then inspect the class of `wf`:
+
+```{r}
+wf %>% class() # class before
+workflows::remove_recipe(wf) %>% class() # class after removing recipe using workflows function
```
dajmcdon, line 100:
```suggestion
workflows::remove_recipe(wf) %>% class() # class after removing recipe using workflows function
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+
+## Adjust a single step of an `epi_recipe`
+
+Suppose that we just want to change a single step in an `epi_recipe` (that is either standalone or a part of an `epi_workflow`). Instead of replacing an entire `epi_recipe`, we can use the `adjust_epi_recipe` function. In this function, the step to be adjusted is indicated either the step number or name in the `which_step` parameter. Then, the parameter name and update value must be inputted as `...`.
+
+For instance, suppose that we decide to lead the `death_rate` by 14 days instead of 7. We may adjust this step in `wf` recipe by setting `which_step` to the step number in the order of operations, which can be obtained by inspecting `r2` or the tidy summary of it:
+
+```{r}
+workflows::extract_preprocessor(wf) # step_epi_ahead is the third step in r2
+tidy(workflows::extract_preprocessor(wf)) # tidy tibble summary of r2
```
dajmcdon, line 176:
```suggestion
extract_preprocessor(wf) # step_epi_ahead is the third step in r2
tidy(extract_preprocessor(wf)) # tidy tibble summary of r2
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+```{r}
+wf %>% adjust_epi_recipe(which_step = "step_epi_ahead", ahead = 14) # not overwrite r2 because same result
+```
+
+If there are at least two steps in a recipe that share the same name, specifying the name in `which_step` will throw an error as `adjust_epi_recipe` is not intended to be used to modify multiple steps at once. The way, then, to modify a step that has the same name as another is to indicate what number it is in the ordering of the steps. For example, in `r2` there are two steps named `step_epi_lag` - the first step where we lag the death rate, and the second where we lag the case rate. If we want to modify the lags for the `case_rate` variable, we would specify the step number of 2 in `which_step`. 
+
+```{r}
+wf <- wf %>% adjust_epi_recipe(which_step = 2, lag = c(0, 1, 7, 14, 21))
+
+workflows::extract_preprocessor(wf)
```
dajmcdon, line 192:
```suggestion
extract_preprocessor(wf)
```

vignettes/update.Rmd
```diff
@@ -0,0 +1,254 @@
...
+```{r}
+adjust_epi_recipe(r2, which_step = 2, lag = c(0, 1, 7, 14, 21)) # should be same result as above
+```
+
+Note that when we adjust the `r2` object directly, we are not adjusting the recipe in the `epi_workflow`. That is, if we modify a step in `r2`, the change will not automatically transfer over to `wf`. We would need to modify the recipe in `wf` directly (`adjust_epi_recipe` on `wf`) or update the recipe in `wf` with a new `epi_recipe` that has undergone the adjustment (using `update_epi_recipe`):
+
+```{r}
+r2 <- adjust_epi_recipe(r2, which_step = 2, lag = 0:21)
+
+workflows::extract_preprocessor(wf)
```
dajmcdon, line 206:
```suggestion
extract_preprocessor(wf)
```

dajmcdon (approved):
I made some suggestions for simplicity. Please try to apply the same throughout the vignette (I suspect I got many, but not all). 

Once you're done, you can merge.

vignettes/update.Rmd
rachlobay: I think I'm going to keep the `workflows::` for now because it's a quick and clear way to indicate what functions I used are from that package & I think it's good to lay that out somewhere (because otherwise it can be hard to tell for a beginner which functions are from that package and which are from ours - because it's not exceedingly common).

PR #281, opened 2024-01-11
Check dots empty

### LINKED ISSUE #276: Missing `check_dots_empty()` in `arx_args_list()`
Probably should be [here](https://github.com/cmu-delphi/epipredict/blob/378577a213aa59043dea73e8fdea725432f32f16/R/arx_forecaster.R#L224), mirroring what classifier args list does [here](https://github.com/cmu-delphi/epipredict/blob/378577a213aa59043dea73e8fdea725432f32f16/R/arx_classifier.R#L255).

Context: Ran into this migrating epipredict versions with the `levels` -> `quantile_levels` rename.

dsweber2:
Closes #276. the different commits identify confidence in actually using `check_dots_used`. The slather cases were the most ambiguous to me

### DIFF
--- a/R/epi_recipe.R
+++ b/R/epi_recipe.R
@@ -57,6 +57,7 @@ epi_recipe.default <- function(x, ...) {
 #' r
 epi_recipe.epi_df <-
   function(x, formula = NULL, ..., vars = NULL, roles = NULL) {
+    rlang::check_dots_empty()
     if (!is.null(formula)) {
       if (!is.null(vars)) {
         rlang::abort(
@@ -288,6 +289,7 @@ is_epi_recipe <- function(x) {
 #'
 add_epi_recipe <- function(
     x, recipe, ..., blueprint = default_epi_recipe_blueprint()) {
+  rlang::check_dots_empty()
   workflows::add_recipe(x, recipe, ..., blueprint = blueprint)
 }
 
--- a/R/epi_recipe.R
+++ b/R/epi_recipe.R
@@ -57,7 +57,6 @@ epi_recipe.default <- function(x, ...) {
 #' r
 epi_recipe.epi_df <-
   function(x, formula = NULL, ..., vars = NULL, roles = NULL) {
-    rlang::check_dots_empty()
     if (!is.null(formula)) {
       if (!is.null(vars)) {
         rlang::abort(
@@ -289,7 +288,6 @@ is_epi_recipe <- function(x) {
 #'
 add_epi_recipe <- function(
     x, recipe, ..., blueprint = default_epi_recipe_blueprint()) {
-  rlang::check_dots_empty()
   workflows::add_recipe(x, recipe, ..., blueprint = blueprint)
 }
 
--- a/R/epi_workflow.R
+++ b/R/epi_workflow.R
@@ -113,6 +113,7 @@ is_epi_workflow <- function(x) {
 #' wf
 #' @export
 add_model <- function(x, spec, ..., formula = NULL) {
+  rlang::check_dots_empty()
   UseMethod("add_model")
 }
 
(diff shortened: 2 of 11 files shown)


R/epi_recipe.R
```diff
@@ -288,6 +289,7 @@ is_epi_recipe <- function(x) {
 #'
 add_epi_recipe <- function(
     x, recipe, ..., blueprint = default_epi_recipe_blueprint()) {
+  rlang::check_dots_empty()
```
dajmcdon, line 292: This one is fine because `add_recipe()` [checks](https://github.com/tidymodels/workflows/blob/b841ff0becf9a4fab4e52ac89e628ca07d015735/R/pre-action-recipe.R#L54)

R/epi_workflow.R
```diff
@@ -113,6 +113,7 @@ is_epi_workflow <- function(x) {
 #' wf
 #' @export
 add_model <- function(x, spec, ..., formula = NULL) {
+  rlang::check_dots_empty()
```
dajmcdon, line 116: This is a generic, so no check.

R/extract.R
```diff
@@ -18,6 +18,7 @@
 #'
 #' extract_argument(f, "layer_residual_quantiles", "symmetrize")
 extract_argument <- function(x, name, arg, ...) {
+  rlang::check_dots_empty()
```
dajmcdon, line 21: another generic, no check.

R/frosting.R
```diff
@@ -303,6 +303,7 @@ frosting <- function(layers = NULL, requirements = NULL) {
 #' @return a `frosting` object
 #' @export
 extract_frosting <- function(x, ...) {
+  rlang::check_dots_empty()
```
dajmcdon, line 306: a generic, no check

R/layer_predict.R
```diff
@@ -43,6 +43,7 @@
 layer_predict <-
   function(frosting, type = NULL, opts = list(), ...,
            id = rand_id("predict_default")) {
+    rlang::check_dots_empty()
```
dajmcdon, line 46: dots are actually used in this function (though not correctly, that's #46

R/tidy.R
```diff
@@ -49,6 +49,7 @@ tidy.frosting <- function(x, number = NA, id = NA, ...) {
   # add id = NA as default. If both ID & number are non-NA, error.
   # If number is NA and ID is not, select the layer with the corresponding
   # ID. Only a single ID is allowed, as this follows the convention for number
+  rlang::check_dots_empty()
```
dajmcdon, line 52: usually [not checked](https://github.com/tidymodels/recipes/blob/36a1307656ea7804eff127fd2294cc1f2dd9421e/R/impute_linear.R#L245)

dajmcdon (changes requested):
Thanks for doing this. The dots arguments are a bit complicated. See for example: https://adv-r.hadley.nz/s3.html#s3-arguments

My understanding is that
1. The S3 generics in this package need `...` because methods may have different arguments. So generics should never `check_dots_*`.
2. methods "can" have `check_dots_*`. However, I've tried to make my versions of functions in `{recipes}` have similar constructions. So, e.g., `epi_recipe` methods don't check because `recipe` methods [don't check](https://github.com/tidymodels/recipes/blob/HEAD/R/recipe.R). I'm not sure I could articulate a reason, but my assumption is "they know better".
3. `add_model()` should check because the `{workflows}` analogues (e.g., [`add_preprocessor()`](https://github.com/tidymodels/workflows/blob/b841ff0becf9a4fab4e52ac89e628ca07d015735/R/workflow.R#L77) does.
4. The `slather()` methods are more complicated. These aren't particularly user facing, so it may not matter. The `bake()` methods in `{recipes}` generally don't check, so I haven't (as in point 2) for example, [`bake.*()/prep.*()`](https://github.com/tidymodels/recipes/blob/36a1307656ea7804eff127fd2294cc1f2dd9421e/R/impute_linear.R#L161)

I've tried to comment on each individually.

dajmcdon (approved):

PR #282, opened 2024-01-18
fix: load_all patch, and Roxygen update

dsweber2:
This makes it so that `quantile_reg` and `smooth_quantile_reg` only get defined as new models if they aren't already in the current model namespace. Debugging these still requires restarting R; for everything else, there's `devtools::load_all()` (which works now).

In the process of writing this, roxygen ran into issues where the new version didn't recognize \`%@%\` and needed just a bare %@%.


R/make_quantile_reg.R
```diff
@@ -49,7 +49,10 @@ quantile_reg <- function(mode = "regression", engine = "rq", quantile_levels = 0
 }
 
 make_quantile_reg <- function() {
-  parsnip::set_new_model("quantile_reg")
+  current_env <- get_model_env()
```
dajmcdon, line 52: Can we do:
`model_env <- parsnip::get_model_env()`

And same below.

R/make_quantile_reg.R
```diff
@@ -49,7 +49,10 @@ quantile_reg <- function(mode = "regression", engine = "rq", quantile_levels = 0
...
+  if (!any(current_env$models == "quantile_reg")) {
```
dajmcdon, line 53: Perhaps `!("quantile_reg" %in% model_env$models)`. I somehow feel like that's clearer?

dajmcdon (approved):
I think you should also delete `.Rprofile`. Once you make these minor changes, feel free to merge.

R/make_quantile_reg.R
dsweber2: sure, that seems a reasonable style thing, I literally copy-pasted this from `check_model_exists`, which throws an error instead of giving the boolean for........ reasons??

PR #286, opened 2024-01-26
refactor: use checkmate for arg validation

dshemetov:
Related to https://github.com/cmu-delphi/epiprocess/issues/403.

I took the safe approach of changing the internals of the arg functions to use `checkmate`, that way we can rely on the existing unit tests and don't need to change any other epipredict code. The functions look a lot simpler and more unified now. We could eventually get rid of these wrappers and just invoke check functions directly, but that would take more refactoring work.


tests/testthat/test-arg_is_.R
```diff
@@ -44,7 +44,7 @@ test_that("scalar", {
 test_that("numeric", {
   expect_silent(arg_is_numeric(i, j, x, y))
   expect_error(arg_is_numeric(a))
-  expect_error(arg_is_numeric(d))
+  expect_silent(arg_is_numeric(d))
```
dshemetov, line 47: Dates now pass `arg_is_numeric`, this didn't affect the rest of the tests.

tests/testthat/test-arg_is_.R
```diff
@@ -68,7 +68,7 @@ test_that("positive", {
 test_that("nonneg", {
   expect_silent(arg_is_nonneg(i, j, x, y))
   expect_error(arg_is_nonneg(a))
-  expect_error(arg_is_nonneg(d))
+  expect_silent(arg_is_nonneg(d))
```
dshemetov, line 71: Dates now pass `arg_is_nonneg`, this didn't affect the rest of the tests.

tests/testthat/test-arg_is_.R
```diff
@@ -96,7 +96,8 @@ test_that("date", {
   expect_error(arg_is_date(d, dd, n))
   expect_error(arg_is_date(d, dd, nn))
   expect_silent(arg_is_date(d, dd, n, allow_null = TRUE))
-  expect_silent(arg_is_date(d, dd, nn, allow_na = TRUE))
+  # Upstream issue, see: https://github.com/mllg/checkmate/issues/256
+  # expect_silent(arg_is_date(d, dd, nn, allow_na = TRUE))
```
dshemetov, line 100: I had to remove `allow_na` from `arg_is_date`, due to a checkmate inconsistency, but this didn't affect the rest of the tests.

tests/testthat/test-arg_is_.R
```diff
@@ -136,14 +137,6 @@ test_that("function", {
...
 })
 
-test_that("sorted", {
-  expect_silent(arg_is_sorted(a = 1:5, b = 6:10))
-  expect_error(arg_is_sorted(a = 5:1, b = 6:10))
-  expect_error(arg_is_sorted(b = NULL))
-  expect_silent(arg_is_sorted(b = NULL, allow_null = TRUE))
-})
-
-
```
dshemetov, line 146: These weren't used anywhere and checkmate supports a sorted check for some data types, so I figured we could rely on that when we need it.

R/epipredict-package.R
```diff
@@ -4,6 +4,10 @@
 #' @importFrom stats poly predict lm residuals quantile
 #' @importFrom cli cli_abort
 #' @importFrom lifecycle deprecated
+#' @importFrom checkmate assert assert_character assert_int assert_scalar
+#'             assert_logical assert_numeric assert_number assert_integer
+#'             assert_integerish assert_date assert_function
+#' @importFrom purrr walk
```
dajmcdon, line 10: We shouldn't need this (or `{purrr}` at all). See https://github.com/cmu-delphi/epipredict/blob/d46e0c5c42eaa2c76ae495ad709bcd62eca5e633/R/compat-purrr.R#L9

dajmcdon (changes requested):
Do these changes result in reasonable/interpretable error messages (still get the argument names correct in the error message, etc.) Can you maybe show an example in the PR comments?

dshemetov:
Good question right, I brought back the `handle_arg_list` since it handled grabbing names well. Checkmate does quite well with preserving variable names in errors.

```r
r$> r1 <- epi_recipe(x) %>%
          step_epi_ahead(death_rate, ahead = 3.6) %>%
          step_epi_lag(death_rate, lag = 1.9)
Error in (function (name, value)  : 
  Assertion on 'ahead' failed: Must be of type 'integerish', but element 1 is not close to an integer.

r$> r2 <- epi_recipe(x) %>%
          step_epi_ahead(death_rate, ahead = 7) %>%
          step_epi_lag(death_rate, lag = -7)
Error in (function (name, value)  : 
  Assertion on 'lag' failed: Element 1 is not >= 0.

r$> step_growth_rate(r, value, role = 1)
Error in (function (name, value)  : 
  Assertion on 'role' failed: Must be of type 'character', not 'double'.

r$> step_growth_rate(r, value, horizon = 0)
Error in (function (name, value)  : 
  Assertion on 'horizon' failed: Element 1 is not >= 1.
```

dshemetov:
FWIW, `traceback()` can lead the user there

```r
r$> r1 <- epi_recipe(x) %>%
              step_epi_ahead(death_rate, ahead = 3.6) %>%
              step_epi_lag(death_rate, lag = 1.9)
Error in (function (name, value)  : 
  Assertion on 'ahead' failed: Must be of type 'integerish', but element 1 is not close to an integer.

r$> traceback()
15: stop(simpleError(sprintf(msg, ...), call.))
14: mstop("Assertion on '%s' failed: %s.", var.name, res, call. = sys.call(-2L))
13: makeAssertion(x, res, .var.name, add)
12: assert_integerish(value, null.ok = allow_null, lower = 0, any.missing = FALSE, 
        .var.name = name) at utils-arg.R#63
11: (function (name, value) 
    {
        assert_integerish(value, null.ok = allow_null, lower = 0, 
            any.missing = FALSE, .var.name = name)
    })(dots[[1L]][[1L]], dots[[2L]][[1L]])
10: mapply(.f, .x, .y, MoreArgs = list(...), SIMPLIFY = FALSE) at compat-purrr.R#72
9: map2(.x, .y, .f, ...) at compat-purrr.R#15
8: walk2(names, values, .tests) at utils-arg.R#10
7: handle_arg_list(..., .tests = function(name, value) {
       assert_integerish(value, null.ok = allow_null, lower = 0, 
           any.missing = FALSE, .var.name = name)
   }) at utils-arg.R#62
6: arg_is_nonneg_int(ahead) at step_epi_shift.R#126
5: step_epi_ahead(., death_rate, ahead = 3.6)
4: inherits(x, "epi_recipe") at epi_recipe.R#221
3: is_epi_recipe(recipe) at step_epi_shift.R#64
2: step_epi_lag(., death_rate, lag = 1.9)
1: epi_recipe(x) %>% step_epi_ahead(death_rate, ahead = 3.6) %>% 
       step_epi_lag(death_rate, lag = 1.9)
```

dajmcdon:
True, though it would be nice if `Error in (function (name, value)  : ` said `Error in step_epi_ahead() :` instead.

PR #294, opened 2024-02-20
Djm/resids hotfix

dajmcdon:
### Checklist

Please:

-   [x] Request a review from one of the current epipredict main reviewers:
        dajmcdon.
-   [x] Makes sure to bump the version number in `DESCRIPTION` and `NEWS.md`.

If `grab_residuals()` returns a data frame with some keys, then the `by_key` argument would cause an error. This is be cause, for example, if `geo_value` was a key (it always is), and the `r` object had the columns `c("geo_value", ".resid")`, then `bind_cols(key_cols, r)` renames the two geo_value columns to be distinct, and subsequent group_by operations failed.

-   Resolves #{issue number}

### DIFF
--- a/tests/testthat/test-dist_quantiles.R
+++ b/tests/testthat/test-dist_quantiles.R
@@ -10,14 +10,6 @@ test_that("constructor returns reasonable quantiles", {
   expect_error(new_quantiles(c(1, 2, 3), c(.1, .2, 3)))
 })
 
-test_that("tail functions give reasonable output", {
-  expect_equal(norm_q_par(qnorm(c(.75, .5), 10, 5)), list(m = 10, s = 5))
-  expect_equal(norm_q_par(qnorm(c(.25, .5), 10, 5)), list(m = 10, s = 5))
-  expect_equal(norm_q_par(qnorm(c(.25, .5), 0, 1)), list(m = 0, s = 1))
-  expect_equal(exp_q_par(qlaplace(c(.75, .5), 10, 5)), list(m = 10, s = 5))
-  expect_equal(exp_q_par(qlaplace(c(.25, .5), 10, 5)), list(m = 10, s = 5))
-  expect_equal(exp_q_par(qlaplace(c(.25, .5), 0, 1)), list(m = 0, s = 1))
-})
 
 test_that("single dist_quantiles works, quantiles are accessible", {
   z <- new_quantiles(values = 1:5, quantile_levels = c(.2, .4, .5, .6, .8))
--- a/tests/testthat/test-dist_quantiles.R
+++ b/tests/testthat/test-dist_quantiles.R
@@ -49,6 +49,34 @@ test_that("quantile extrapolator works", {
   expect_length(parameters(qq[1])$q[[1]], 7L)
 })
 
+test_that("small deviations of quantile requests work", {
+  l <- c(.05, .1, .25, .75, .9, .95)
+  v <- c(0.0890306, 0.1424997, 0.1971793, 0.2850978, 0.3832912, 0.4240479)
+  badl <- l
+  badl[1] <- badl[1] - 1e-14
+  distn <- dist_quantiles(list(v), list(l))
+
+  # was broken before, now works
+  expect_equal(quantile(distn, l), quantile(distn, badl))
+
+  # The tail extrapolation was still poor. It needs to _always_ use
+  # the smallest (largest) values or we could end up unsorted
+  l <- 1:9 / 10
+  v <- 1:9
+  distn <- dist_quantiles(list(v), list(l))
+  expect_equal(quantile(distn, c(.25, .75)), list(c(2.5, 7.5)))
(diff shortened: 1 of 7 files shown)


dsweber2:
Interesting, a little surprised I haven't actually run into this problem. If we're not doing grouping by `geo_value` by default, when would we want to do this? Is this a problem that comes up naturally for multi-aheads, e.g. `smooth_quantile_regression`?

So I tried the test without the change and it seems to have passed; doing a `browser()`, it looks like `grab_residuals` only returns a `.resid` column. My guess is that you had a more complicated situation where `grab_residuals` actually had too many columns that made this unit test from?

dsweber2 (changes requested):
Probably need to extract the test case slightly differently. Looking at `grab_residuals`, seems like it could only happen if `".resid"` is already in the residuals of `the_fit`, which is returning a dataframe.

Were you maybe using a different engine than `linear_reg`?

dajmcdon:
Here's the failure:

``` r
library(epipredict)
#> Loading required package: epiprocess
#> 
#> 
#>     filter
flat1 <- flatline_forecaster(case_death_rate_subset, "death_rate")
flat2 <- flatline_forecaster(
  case_death_rate_subset, "death_rate",
  args_list = flatline_args_list(quantile_by_key = "geo_value")
)
#> New names:
#> • `geo_value` -> `geo_value...1`
#> • `geo_value` -> `geo_value...2`
#> Error in `dplyr::group_by()`:
#> ! Must group by variables found in `.data`.
#> ✖ Column `geo_value` is not found.
```

<sup>Created on 2024-02-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


And on rebuilding with this fix:

``` r
library(epipredict)
#> Loading required package: epiprocess
#> 
#> 
#>     filter
flat1 <- flatline_forecaster(case_death_rate_subset, "death_rate")
flat2 <- flatline_forecaster(
  case_death_rate_subset, "death_rate",
  args_list = flatline_args_list(quantile_by_key = "geo_value")
)
```

<sup>Created on 2024-02-20 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

dajmcdon:
It's because of the `grab_residuals()` function returning a tibble with 2 columns (`geo_value` and `.resid`) for the flatline engine. This doesn't (necessarily) happen with other engines.

dsweber2:
That seems simple enough to include as the unit test maybe?

I tried the straightforward `wf <- epi_workflow(r, parsnip::linear_reg() %>% parsnip::set_engine("flatline")) %>% fit(jhu)` to get the test to fail on the old version, but that gives the confusing error of
```
Error in `dplyr::arrange()`:
ℹ In argument: `..1 = time_value`.
Caused by error:
! object 'time_value' not found
```

tests/testthat/test-dist_quantiles.R
```diff
@@ -49,6 +41,34 @@ test_that("quantile extrapolator works", {
...
+
+  # was broken before, now works
+  expect_equal(quantile(distn, l), quantile(distn, badl))
+
+  # The tail extrapolation was still poor. It needs to _always_ use
+  # the smallest (largest) values or we could end up unsorted
+  l <- 1:9 / 10
+  v <- 1:9
+  distn <- dist_quantiles(list(v), list(l))
+  expect_equal(quantile(distn, c(.25, .75)), list(c(2.5, 7.5)))
```
dsweber2, line 59: Sanity check, this is 2.5 because that's halfway between 2 and 3, which are the `.2` and `.3` quantile values?

tests/testthat/test-dist_quantiles.R
```diff
@@ -49,6 +41,34 @@ test_that("quantile extrapolator works", {
...
+  expect_equal(quantile(distn, c(.1, .9)), list(c(1, 9)))
+  qv <- data.frame(q = l, v = v)
+  expect_equal(
+    unlist(quantile(distn, c(.01, .05))),
+    tail_extrapolate(c(.01, .05), head(qv, 2))
+  )
```
dsweber2, line 65: Not sure I quite get how this is testing the tail behavior; is it just checking that `quantile` is using `tail_extrapolate` to calculate the values at those quantiles?

dsweber2 (approved):
Half of this is addressing a new issue right? Seems like you changed the extrapolation to use the same method regardless of region and refactored some functions?

I have lingering questions about things but I wouldn't let those block

dsweber2:
oh, I pushed the change for the styler, which seems to have been the only thing that was breaking the CI

dajmcdon:
@dsweber2 Am I good to merge this?

dsweber2:
oh, I suppose I should've said that in addition to approving. Would you prefer I just merge if I approve when I'm tagged?

But yes, lgtm!

dajmcdon:
Ah, no problem. I think better for the PR creator to merge once satisfied, but I lost track of whether we were both satisfied!

PR #295, opened 2024-03-06
Djm/plotting

dajmcdon:
### Checklist

Please:

-   [ ] Request a review from one of the current epipredict main reviewers:
        dajmcdon.
-   [x] Makes sure to bump the version number in `DESCRIPTION` and `NEWS.md`.

-   Resolves #{issue number}


DESCRIPTION
```diff
@@ -51,7 +52,6 @@ Suggests:
     data.table,
     epidatr (>= 1.0.0),
     fs,
```
dshemetov, line 54: Did `fs` get accidentally added back in after the "prune imports" merge?

dajmcdon: No this is correct. It should be in Suggests (got pruned from Imports).

dshemetov (approved):
Made a few docs change and added a few snapshot tests. Example plots looked good locally.

PR #296, opened 2024-03-18
Adjust ahead

### LINKED ISSUE #279: Add latency adjustment
See the implementation here: https://github.com/cmu-delphi/exploration-tooling/blob/77d7e5eb95e4b17f40567430b0233f6b24cf100a/R/latency_adjusting.R#L11.

* One possibility is as an argument to `step_epi_ahead()`. 
    - pros: easier adjustment in current canned methods. Just add an argument to all the canned steps and proceed.
    - cons: limits the flexibility to simply shifting based on explicit missingness (or carefully specified `target_date`). If the `target_date` is autocalculated based on the `as_of` for the `epi_df`, the ahead could shift massively (say when using finalized data).
* The other is as a separate (new) step.
    - pros: much more flexibility, if needed (do we?)
    - cons: more difficult to propagate through downstream. 

Deliverables:
- [ ] A new / adjusted step.
- [ ] Propagate throughout examples / vignettes where necessary.
- [ ] Check that all canned forecaster workflows operate appropriately.
- [ ] Add a vignette describing the behaviour and alternatives (LOCF implementation, imputing using steps in `{recipes}`)
- [ ] Adjust `get_test_data()` as needed

dsweber2:
### Checklist

Please:

-   [X] Make sure this PR is against "dev", not "main".
-   [X] Request a review from one of the current epipredict main reviewers:
        dajmcdon.
-   [X] Makes sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
-   [X] Describe changes made in NEWS.md, making sure breaking changes

Implementation of latency adjustments as an argument to `step_epi_ahead`. The rough idea is to make `ahead` relative to the `as_of` date, rather than the last day of data. It throws a warning if this is too far in the future. Only `extend_ahead` and `None` are currently implemented. Biggest downside is it only works for the `epi_df` that the pipeline was created on.

Still working on:
- [x] some tests (such as making sure it only works on the first `epi_df` and the end-to-end application works, rather than just the specialized function `extend_ahead`).
- [x] adjustments to layers to account for the existence of `step_adjust_ahead`
- [x] `locf` and associated tests
- [x] `extend_lags` we should talk about this one, it's a new-ish @lcbrooks idea. Sort of the opposite of `extend_aheads`. Setting the forecast date to be `as_of`, per predictor, adjust the lags so that they're relative to the last day of data. E.g. if predictor `A` has 3 days of latency, and the lags are given as `(0,7,14)`, set the lags to `(3,10,17)`, while if `B` has 5 days of latency, set the lags for `B` to `(5,12,19)`. Feel free to move discussion to the issue
- [x] demo vignettes

-   Resolves #279


R/utils-shift.R
```diff
@@ -0,0 +1,71 @@
...
+        ))
+      }
+      effective_ahead <- as.integer(shift_amount + ahead)
+      time_type <- attributes(new_data)$metadata$time_type
+
+      if ((grepl("day", time_type) && (shift_amount >= 10)) ||
+        (grepl("week", time_type) && (shift_amount >= 4)) ||
+        ((time_type == "yearmonth") && (shift_amount >= 2)) ||
+        ((time_type == "yearquarter") && (shift_amount >= 1)) ||
+        ((time_type == "year") && (shift_amount >= 1))) {
```
dsweber2, line 49: The logic for when to do this warning could use some input. My initial thought was the ratio between shift_amount/ahead, but comparing ahead=1 and ahead=10, there's no really reasonable way to allow 1 -> 5 but not allow 10 -> 50 using the ratio.

The above is very much a guess, especially for time_types other than day

dsweber2:
based on feedback, I'm making this a separate step and making sure full-pipeline tests work. It will be a PR until both of those are done.

dsweber2:
So I'm still putting the tests in a format that isn't a pile of spaghetti, but I wanted to get this out so we could maybe talk about it. 

To implement a separate step, I couldn't find a way to get previous steps, so I basically had to communicate everything relevant through the headers of `new_data` in the bake step.  Hopefully modifying those doesn't cause problems.

While I was looking around I noticed that both `step_epi_lag` and `step_epi_ahead` were essentially the same; I made them call a separate function to highlight the actual differences. Couldn't quite reuse that for `step_adjust_latency` unfortunately, though that has a similar thing where `adjust_ahead` and `adjust_lag` basically do the same thing.

There's still some downstream implications for how `layer_forecast_date` works, specifically in the default case. We may have to add it as an option here too for it to work for the case of someone setting it manually using that step, which seems a bit unfortunate but kinda unavoidable

R/utils-latency.R
```diff
@@ -0,0 +1,171 @@
...
+    pmap(\(original_name, latency, new_name) {
+      epi_shift_single(
+        x = new_data,
+        col = original_name,
+        shift_val = latency,
+        newname = new_name,
+        key_cols = keys
+      )
+    }) %>%
+    map(\(x) zoo::na.trim(x)) %>% # TODO need to talk about this
```
dsweber2, line 23: Two things about this:
1. Zoo dependency. Seems we don't currently have this. are we trying to minimize package dependencies? This is the only function I use from it
2. This shears `NA`'s off the end of the column. We have a step that explicitly drops `NA`'s, so I'm not sure how much we're trying to avoid removing `NA`'s otherwise. Not sure how to do this correctly otherwise though.

dsweber2:
Getting a peculiar error where the local snapshot test for population scaling has 4 lines like this
```R
    Code
      p <- predict(wf, latest)
    Message
      Joining with `by = join_by(geo_value)`
      Joining with `by = join_by(geo_value)`
      Joining with `by = join_by(geo_value)`
      Joining with `by = join_by(geo_value)`
```
whereas the CI is apparently only generating 2. And this apparently changed within the past 2 weeks.

dajmcdon:
This was a change in `{dplyr}` at some point that generates these if you don't specify `by =`. I tried purposely to get rid of them. But maybe I missed some? Or your new code is recreating more?

dajmcdon:
A few other notes (we can discuss synchronously):

1. The only time a `step_*()` function has access to the recipe is when it is added. At that point, it can "see" all previous steps. So inside of `step_adjust_latency()`, you can immediately check to see if there are leads/lags in the current set of recipe instructions. In principle, you could then adjust them if desired. I'm not sure this handles everything, but I suspect it handles some of it.
2. The `map(\(x), x+3)` style syntax was added in R 4.1. We've avoided forcing this Dependency on users to this point. I'm open to changing this (it's almost 3 years old), but we should discuss. This is also the reason for sticking with the `{magrittr}` pipe `%>%` rather than the native pipe `|>`.
3. The `{zoo}` dependency. There are situations where removal is _always_ the right thing (creating NAs on future `time_value`'s is probably one of these), and our `step_epi_lead()` does this currently. However, my previous practice has been to pass any NAs created by steps along. This way, additional steps can potentially be added to handle them (say, any of [these](https://recipes.tidymodels.org/reference/index.html#step-functions-imputation), or similar time series fill versions). Simplest is the `step_naomit()` functions. Is there a reason that this can't be done without the dependency or handled downstream?

tests/testthat/test-population_scaling.R
```diff
@@ -207,7 +207,63 @@ test_that("Postprocessing to get cases from case rate", {
...
+      suffix = "_scaled"
+    ) %>%
+    step_epi_lag(case_rate_scaled, lag = c(0, 7, 14)) %>% # cases
+    step_epi_ahead(case_rate_scaled, ahead = 7, role = "outcome") %>% # cases
+    recipes::step_naomit(recipes::all_predictors()) %>%
+    recipes::step_naomit(recipes::all_outcomes(), skip = TRUE)
+
+  expect_snapshot(prep <- prep(r, jhu))
+
+  expect_snapshot(b <- bake(prep, jhu))
```
dsweber2, line 233: So these are the tests generating the snapshots. Looking at my local timemachine, this was a set of tests that was previously being skipped, and I added these as snapshot tests. I guess I'll just put this back to skipped for now.

dsweber2: But yes, it does look like there are some lingering `Joining with `by = join_by(geo_value)` messages

R/utils-latency.R
```diff
@@ -0,0 +1,191 @@
...
+      current_column_names = names(new_data),
+      class = "epipredict_adjust_latency_nonexistent_column_used",
+      call = call
+    )
+  }
+  # this pulls text that is any number of digits between two _, e.g. _3557_, and
+  # converts them to an integer
+  shift_amounts <- stringr::str_match(relevant_columns, "_(\\d+)_") %>%
+    `[`(, 2) %>%
+    as.integer()
```
dajmcdon, line 70: I'm a bit concerned about the robustness of this. For example, if I did `step_epi_lag(..., prefix = "la_g_",...)`, this would fail. I think we need to examine the recipe (then you can just capture the lags/leads directly).

R/utils-latency.R
```diff
@@ -0,0 +1,191 @@
...
+
+  shift_cols <- dplyr::tibble(
+    original_name = relevant_columns,
+    shifts = shift_amounts
+  )
+  if (is.null(latency)) {
+    shift_cols %<>%
```
dajmcdon, line 77: I'd like to avoid this nonstandard pipe. I like it, but this can be confusing to most R users.

R/utils-latency.R
```diff
@@ -0,0 +1,191 @@
...
+    mutate(
+      effective_shift = shifts + abs(latency)
+    ) %>%
+    mutate(
+      new_name = adjust_name(prefix, original_name, effective_shift)
+    )
+  info %<>% select(variable, type, role)
+  shift_cols <- left_join(shift_cols, info, by = join_by(original_name == variable))
+  if (length(unique(shift_cols$role)) != 1) {
+    cli::cli_error("not all roles are the same!",
```
dajmcdon, line 106: I don't see this function in the package. Maybe `cli::cli_abort()`?

R/utils-latency.R
```diff
@@ -0,0 +1,191 @@
...
+  as_of <- attributes(new_data)$metadata$as_of
+  max_time <- max(time_values)
+  # make sure the as_of is sane
+  if (!inherits(as_of, class(time_values))) {
+    rlang::abort(glue::glue(
+      "the data matrix `as_of` value is {as_of}, ",
+      "and not a valid `time_type` with type ",
+      "matching `time_value`'s type of ",
+      "{typeof(new_data$time_value)}."
+    ))
```
dajmcdon, line 146: Is there a reason for rlang versions here and cli elsewhere?

R/utils-latency.R
```diff
@@ -0,0 +1,191 @@
...
+#' @keywords internal
+#' @importFrom stringi stri_replace_all_regex
+adjust_name <- function(prefix, column, effective_shift) {
+  pattern <- paste0(prefix, "\\d+", "_")
+  adjusted_shifts <- paste0(prefix, effective_shift, "_")
+  stringi::stri_replace_all_regex(
+    column,
+    pattern, adjusted_shifts
+  )
+}
```
dajmcdon, line 179: Probably a similar issue as above.

R/utils-shift.R
```diff
@@ -0,0 +1,71 @@
...
+  if (is.na(method) || is.null(method) || method == "None") {
+    return(object$ahead)
+  } else if (method == "extend_ahead") {
+    as_of <- attributes(new_data)$metadata$as_of
+    if (FALSE && (typeof(as_of) != typeof(new_data$time_value))) {
+      rlang::abort(glue::glue(
+        "the data matrix `as_of` value is {as_of}, ",
+        "and not a valid `time_type` with type ",
+        "matching `time_value`'s type of ",
+        "{typeof(new_data$time_value)}."
```
dajmcdon, line 19: `cli`?

dajmcdon (changes requested):
I left a number of questions and comments through out. It may be easiest to schedule a check in or maybe on Slack?

R/utils-latency.R
brookslogan: - by default, `na.trim` removes from both the beginning and the end
- side note: `\(` is R >= 4.1.  We were recently talking about not trying to support versions 3.* but not sure about 4.0.
- this function from latency processing code lying around may be helpful:
```r
n_true_at_tail = function(x) {
  match(FALSE, rev(x), nomatch = length(x) + 1L) - 1L
}
# So then to remove NAs at tail you could do the following (or maybe an S3 version of this):
remove_na_tail <- function(x) {
  x[seq_len(length(x) - n_true_at_tail(is.na(x)))]
}

remove_na_tail(c(NA, NA, 1:10, NA, 1:10, NA, NA))
remove_na_tail(1:100) # ALTREP
remove_na_tail(1:100 + 1) # non-ALTREP
remove_na_tail(integer(0L))
remove_na_tail(rep(NA_integer_, 5))
```

R/utils-latency.R
dsweber2: nope, not sure I put much thought into it. The package is also in a split state; is it alright if I just find-replace all instances of `rlang::abort` with `cli::cli_abort`, or is there some reason there are both generally?

R/step_adjust_latency.R
```diff
@@ -0,0 +1,273 @@
...
+    if (!is_epi_recipe(recipe)) {
+      rlang::abort("This recipe step can only operate on an `epi_recipe`.")
+    }
+    if (!is.null(columns)) {
+      rlang::abort(c("The `columns` argument must be `NULL.",
+        i = "Use `tidyselect` methods to choose columns to lag."
+      ))
+    }
+
+    method <- rlang::arg_match(method)
```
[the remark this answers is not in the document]
dsweber2: >    Partial matches trigger an error.
>    Error messages are a bit more informative and obey the tidyverse standards.
both seem valuable to me

R/utils-latency.R
dajmcdon: At the beginning, we used `{rlang}`, but I've been trying to have all new code use `{cli}` (with a plan to find and replace all the old stuff eventually). I thought there was an issue, but I can't find it.

R/step_adjust_latency.R
```diff
@@ -0,0 +1,273 @@
...
+    if (method == "extend_ahead") {
+      prefix <- "ahead_"
+      if (!any(map_lgl(
+        recipe$steps,
+        \(recipe_step) inherits(recipe_step, "step_epi_ahead")
```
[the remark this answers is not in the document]
dsweber2: oh right. it's short enough in this case I'm happy enough to just replace, but I do wonder if e.g. [tidyverse support windows matter for us](https://www.tidyverse.org/blog/2019/04/r-version-support/)

R/utils-latency.R
brookslogan: You can't generally just find-replace.  See [this migration info](https://rlang.r-lib.org/reference/topic-condition-formatting.html#transitioning-from-abort-to-cli-abort-).

R/step_adjust_latency.R
```diff
@@ -0,0 +1,273 @@
...
+    ) {
+      cli::cli_warn(c(
+        "!" = glue::glue(
+          "The shift has been adjusted by {latency}, ",
+          "which is questionable for it's `time_type` of ",
+          "{time_type}"
+        ),
+        "i" = "input ahead: {shift_cols$shifts[[i_latency]]}",
+        "i" = "shifted ahead: {shift_cols$effective_shift[[i_latency]]}",
+        "i" = "max_time = {max_time} -> as_of = {as_of}"
```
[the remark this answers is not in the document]
dsweber2: I've been burned by it multiple times when using my external version of latency adjustment, and the error that came up as a result wasn't very comprehensible. I'm not sure I'd ever use `step_adjust_latency` without also using `check_latency`.

We could also set the levels to something more extreme, e.g. latency > 365 for day, or 48 for week etc, since it was usually having `as_epi_df` defaulting to the `as_of` being today that was causing me problems with retrospective forecasting

R/step_epi_shift.R
```diff
@@ -109,6 +121,12 @@ step_epi_ahead <-
            ahead,
            prefix = "ahead_",
            default = NA,
+           latency_adjustment = c(
+             "None",
```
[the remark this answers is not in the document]
dsweber2: I... forgot to remove this. But also no good reason.

R/step_epi_shift.R
```diff
@@ -20,6 +20,10 @@
 #' @param lag,ahead A vector of integers. Each specified column will
 #'  be the lag or lead for each value in the vector. Lag integers must be
 #'  nonnegative, while ahead integers must be positive.
+#' @param latency_adjustment a character. Determines the method by which the forecast handles data that doesn't extend to the day the forecast is made. The options are:
+#'   - `"extend_ahead"`: actually forecasts from the last date. E.g. if there are 3 days of latency for a 4 day ahead forecast, the ahead used in practice is actually 7.
+#'   - `"locf"`: carries forward the last observed value up to the forecast date.
+#'   - `"extend_lags"`: per `epi_key` and `predictor`, adjusts the lag so that the shortest lag at predict time is
```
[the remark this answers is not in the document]
dsweber2: it shouldn't, I forgot to drop the doc changes when I switched to having a separate step. thanks for the catch

R/utils-latency.R
dsweber2: ah, they have a guide! Thanks Logan. I'll go over the changed ones and make sure they're not broken. Thankfully none of them broke the tests, but I'm guessing we don't have super strict `expect_error` tests for every single one

R/utils-latency.R
dsweber2: at the moment `step_adjust_latency` also has a `prefix` argument, so if you're setting it in `step_epi_lag` you'd also set it in `step_adjust_latency`, and would avoid the problem. I guess if you started choosing prefixes like `lag_3_` so that  there'd be multiple numbers it would break.

I'll look into grabbing the lags/leads at initial construction time again. I think there may be some ambiguity about which previous step to be inheriting from, but that's probably resolvable.

DESCRIPTION
```diff
@@ -43,12 +43,15 @@ Imports:
...
+    stringr,
+    stringi,
     tibble,
     tidyr,
     tidyselect,
     usethis,
     vctrs,
-    workflows (>= 1.0.0)
+    workflows (>= 1.0.0),
+    zoo
```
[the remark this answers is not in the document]
dsweber2: `zoo` I needed to remove the potential extra `NA`'s that I introduced by the extra shifting. Not sure whether we want to leave that to `step_epi_naomit`. If there are in fact `NA`'s at the end in the original dataset, this will also trim those out.

DESCRIPTION
```diff
@@ -43,12 +43,15 @@ Imports:
     rlang,
     smoothqr,
     stats,
+    stringr,
+    stringi,
```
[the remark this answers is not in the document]
dsweber2: brief reason? I'll try to remove them regardless, I'm just wondering

R/utils-latency.R
dsweber2: right, the problem is that `recipes_eval_select`/ the actual `terms` isn't available during initial construction, so I don't actually know which columns the user is trying to get at. And `prep` doesn't know anything about the preceding functions.

DESCRIPTION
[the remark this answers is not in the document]
dajmcdon: Brief: "fewer Imports is better"
Longer: `{stringi}` uses some beefy compiled code (even the precompiled binaries are large), and can sometimes fail to install on some systems. There are a handful of packages like this, `{sf}` is another, that I prefer to avoid, where possible.

R/utils-latency.R
dsweber2: > [oops just repeating Daniel here] 

yeah, threading in github issues can be hard to follow, especially when there are a lot of comments >.<

could you add some docs to those functions?

dsweber2:
Overview of new changes:
1. I dropped `stringr` and `stringi`, added `purrr`.
2. `zoo` is still present, waiting for us to decide on whether we want to trim or not.
3. I moved lag/ahead detection into the step creation thanks to liberal use of `recipes_eval_select`. If the template is different from the training data this could cause problems, but I guess that's supposed to be true anyways.

R/epi_workflow.R
```diff
@@ -187,7 +187,7 @@ augment.epi_workflow <- function(x, new_data, ...) {
   if (epiprocess::is_epi_df(predictions)) {
     join_by <- epi_keys(predictions)
   } else {
-    rlang::abort(
+    cli::cli_abort(
       c(
         "Cannot determine how to join new_data with the predictions.",
         "Try converting new_data to an epi_df with `as_epi_df(new_data)`."
```
dajmcdon, line 193:
```suggestion
    cli::cli_abort(
      c(
        "Cannot determine how to join `new_data` with the `predictions`.",
        "Try converting `new_data` to an {.cls epi_df} with `as_epi_df(new_data)`."
```

R/step_adjust_latency.R
```diff
@@ -0,0 +1,274 @@
...
+           ),
+           fixed_latency = NULL,
+           fixed_asof = NULL,
+           default = NA,
+           skip = FALSE,
+           columns = NULL,
+           id = recipes::rand_id("epi_lag")) {
+    arg_is_chr_scalar(id, method)
+    if (!is_epi_recipe(recipe)) {
+      cli::cli_abort("This recipe step can only operate on an `epi_recipe`.")
```
dajmcdon, line 94:
```suggestion
      cli::cli_abort("This recipe step can only operate on an {.cls epi_recipe}.")
```

R/step_adjust_latency.R
```diff
@@ -0,0 +1,274 @@
...
+      rel_step_type <- "step_epi_lag"
+      shift_name <- "lag"
+    }
+    relevant_shifts <- construct_shift_tibble(terms_used, recipe, rel_step_type, shift_name)
+
+    if (!any(map_lgl(
+      recipe$steps,
+      function(recipe_step) inherits(recipe_step, rel_step_type)
+    ))) {
+      cli:cli_abort("there is no `{rel_step_type}` defined before this. for the method `extend_{shift_name}` of `step_adjust_latency`, at least one {shift_name} must be previously defined.")
```
dajmcdon, line 122:
```suggestion
      cli:cli_abort(c("There is no `{rel_step_type}` defined before this.",
         "For the method `extend_{shift_name}` of `step_adjust_latency()`,",
         "at least one {shift_name} must be previously defined."
        ))
```

R/step_adjust_latency.R
```diff
@@ -0,0 +1,274 @@
...
+    )
+  }
+
+# lags introduces max(lags) NA's after the max_time_value.
+# TODO all of the shifting happens before NA removal, which saves all the data I might possibly want; I should probably add a bit that makes sure this operation is happening before NA removal so data doesn't get dropped
+#' @export
+prep.step_adjust_latency <- function(x, training, info = NULL, ...) {
+  if ((x$method == "extend_ahead") && (!("outcome" %in% info$role))) {
+    cli::cli_abort('If `method` is `"extend_ahead"`, then a step ",
+          "must have already added an outcome .')
```
dajmcdon, line 168:
```suggestion
cli::cli_abort(c(
  "If `method` is {.val extend_ahead}, then a step ",
  "must have already added an outcome."
))
```

R/step_adjust_latency.R
```diff
@@ -0,0 +1,274 @@
...
+  } else if (!("predictor" %in% info$role)) {
+    cli::cli_abort('If `method` is `"extend_lags"` or `"locf"`, then a step ",
```
dajmcdon, line 170:
```suggestion
cli::cli_abort(c(
  "If `method` is {.val extend_lags} or {.val locf}, then a step ",
  "must have already added a predictor."
))
```

R/step_growth_rate.R
```diff
@@ -56,13 +56,13 @@ step_growth_rate <-
            id = rand_id("growth_rate"),
            additional_gr_args_list = list()) {
     if (!is_epi_recipe(recipe)) {
-      rlang::abort("This recipe step can only operate on an `epi_recipe`.")
+      cli::cli_abort("This recipe step can only operate on an `epi_recipe`.")
```
dajmcdon, line 59:
```suggestion
      cli::cli_abort("This recipe step can only operate on an {.cls epi_recipe}.")
```

R/step_growth_rate.R
```diff
@@ -56,13 +56,13 @@ step_growth_rate <-
...
     }
     method <- match.arg(method)
     arg_is_pos_int(horizon)
     arg_is_scalar(horizon)
     if (!is.null(replace_Inf)) {
-      if (length(replace_Inf) != 1L) rlang::abort("replace_Inf must be a scalar.")
+      if (length(replace_Inf) != 1L) cli::cli_abort("replace_Inf must be a scalar.")
```
dajmcdon, line 65:
```suggestion
      if (length(replace_Inf) != 1L) cli::cli_abort("`replace_Inf` must be a scalar.")
```

R/step_lag_difference.R
```diff
@@ -33,16 +33,16 @@ step_lag_difference <-
            skip = FALSE,
            id = rand_id("lag_diff")) {
     if (!is_epi_recipe(recipe)) {
-      rlang::abort("This recipe step can only operate on an `epi_recipe`.")
+      cli::cli_abort("This recipe step can only operate on an `epi_recipe`.")
```
dajmcdon, line 36:
```suggestion
      cli::cli_abort("This recipe step can only operate on an {.cls epi_recipe}.")
```

tests/testthat/test-epi_shift.R
dajmcdon, line 1: Is there a reason to remove these?

R/step_adjust_latency.R
```diff
@@ -0,0 +1,274 @@
...
+#'
+#' @family row operation steps
+#' @rdname step_adjust_latency
+#' @export
+#' @examples
+#' r <- epi_recipe(case_death_rate_subset) %>%
+#'   step_epi_ahead(death_rate, ahead = 7) %>%
+#'   #   step_adjust_latency(method = "extend_ahead") %>%
+#'   step_epi_lag(death_rate, lag = c(0, 7, 14))
+#' r
```
dajmcdon, line 75: It would help me to flesh this out a bit. Maybe something with latency and see the results with and without?

dajmcdon (approved):
Ok. I think everything looks potentially fine. It would help for the example to be a bit more elaborate. 

Potential for merge conflicts with #331.

tests/testthat/test-epi_shift.R
dsweber2: the function `epi_shift` isn't used anywhere, so I removed it and its tests. Looks like it's an outdated version of `epi_shift_single`

R/step_adjust_latency.R
dsweber2: For now, do the ones in the tests generally address that? I still need to update the layer functions to use dates correctly, but was waiting until these steps are relatively stable. Once I have those, I'll definitely expand the `@examples`, and then write a vignette

R/step_adjust_latency.R
dsweber2: unfortunately this renders as 3 separate lines. Going to need to insert a `glue`. Unless you were suggesting the linebreak?

R/step_adjust_latency.R
dsweber2: what does `{.val` do? in this case I was trying to make a R code string in the docs, like this: `"extend_ahead"`

DESCRIPTION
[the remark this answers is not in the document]
dajmcdon: Let's just defer `NA` removal to the explicit step function.

brookslogan:
Wishlist item based on some dogfooding: if the minimum reporting latency across signals&geos is really large, say, a month or greater, that's worth a warning.  E.g., FluSurv-NET has gaps in reporting between seasons, and I tried to forecast a new season before the first report actually was published using ahead adjustment, and happily tried to forecast the beginning of that season from the end of the preceding season, likely getting poor predictions.

R/utils-shift.R
dsweber2: @lcbrooks I believe this is your wishlist item?

dajmcdon:
@dsweber2 I'm going to make what may seem like an annoying request:

This processing seems sufficiently complicated that I think it should be on it's own, not in `arx_forecaster()`. I don't think that `arx_forecaster()` should accommodate every specialized step/layer that we end up creating. So maybe this needs to just be a new canned forecaster? What do you think?

dsweber2:
I'm not sure it'd be different enough from `arx_forecaster` to have as a completely separate version (comparing `arx_forecaster`, `arx_classifier` and... `arx_latency_adjusted_forecaster`?). What part of the complication pushes you to not include it? Is your concern for having too many options for `arx_forecaster`? Too much complexity to understanding the options? Maintaining the additional feature?

I'd be ok relegating it to just an advanced feature if necessary, though I do think without it, the `forecast_date` and `target_date` arguments are kind of half-baked.

brookslogan:
My 2 cents:
- We want this adjustment for every forecasting Hub task I'm aware of.
- We want this adjustment every time we use multiple predictors unless they are perfectly in sync reporting-latency-wise.

My preference/request would that there be a canned forecaster that includes this feature.  Could be `arx_forecaster()`, could be an `arx_forecaster_from_forecast_date()` (or better-named) alternative.

Side note: I also share some misgivings about the `forecast_date` and `target_date` settings... they always felt like stop-gaps until we had access to this latency/ahead-adjustment feature.  I'm not sure they'd be needed anymore --- in either the max `time_value`-relative or the true-`forecast_date`-relative approaches --- once the adjustment approach is available; presumably, you'd only use the non-adjusted version when you actually want things to be relative to the max `time_value`.

brookslogan:
From our discussion yesterday, realized I overlooked an important possibility of someone newer forecasting without a Hub setup.  Then there is no Hub validator to yell at you for forecasting a changing set of (forecast date relative) aheads when target latency changes / you pull data too early in the day, but you also may be able to say different aheads are fine.  I think maybe the new forecasting examples Richard and Ryan N. are working on may be in that regime still and may give some insight; not sure if we might also have some experiences with epipredict precursors before running things in Hub validator.  Ryan N. already ran into some confusion with (not easily being able to use epi_slide for forecasting and) the epix_slide time_value forecast_date target_date labeling, though the obvious issue there was (typing and recycling rules and) poor epiprocess labeling; I'm not sure if there is also an epipredict thing to worry about there.

Still probably don't want to streamline misrepresenting what the real target dates are in this case.  (And some surprises might come if/when (a) setting forecast date label to be the max time value and that not being the weekday expected, (b) using the true forecast date and target dates as labels and seeing the aheads changing when trying to retrospectively evaluate or plot plus potential uneven spacing or duplicate target date values (per epikey-ahead-forecaster), (c) not using target date and assuming aheads inputted are relative to today in decisionmaking or plotting code and being a bit off, and/or (d) running forecasts too early in the day (stale forecast if AR, not sure what all can happen with ARX, and it's also an issue for the adjusted version but with different behavior).)

dsweber2:
Ok, pretty major revision done, and I think this is ready, modulo fixing some vignettes. 

### The changes that *aren't* features added by adjustAhead:
- adding `smoothqr` to the description, since apparently that was necessary, as well as `grf_quantiles` to the pkgdown.yml
- a few straggling `rlang::abort` -> `cli_abort` conversions, as well as some cases where `c` is used instead of `paste` when the text is meant to continue across the line, rather than form a new bullet point.
- unifying `step_epi_ahead` and `step_epi_lag`'s backend to use the function  `add_shifted_columns`, since they're effectively the same function modulo a sign.
- dropped `epi_shift`, since it was completely unused (and the tests going with it).

### Things tangential to adjustAhead:
- removing any of the completion behavior from `get_test_data` and `forecast` since that kind of `locf` is expressed in a method of `step_adjust_latency`.
- `forecast_date` and `target_date` for `arx_forecaster` are now used in the actual forecast, so if you set them to something so that there's induced latency, but don't select a method, it will warn about the difference. It will also error if `forecast_date`, `target_date`, and `ahead` are all set and are inconsistent.
- `arx_classifier` uses `across(starts_with("ahead"...`, since we don't necessarily know the ahead apriori if they're using `adjust_latency`. Behavior is the same though.
- I added the date of last available data to the print fields of the canned epipreds, along with the adjust amounts (only if some adjust method is set).
- `layer_add_forecast_date`, in the absence of a specific date, inherits the date from `step_adjust_latency` (or if its missing uses `max_time_value`)
- similarly for `layer_add_target_date`

### AdjustAhead dependencies:
- `purrr` dependency, because there was some bug in `map` that the patch was causing problems with.



### Rough overview of the implementation (for arguments and usage, see the rather long man page):
`step_adjust_latency` modifies the `epi_df` so that it has some extra metadata following it around:
1. `latency_table`, a table of how much each column present at the time of calling `step_adjust_latency` is latent relative to the `forecast_date`
2. `metadata`, which is a cached version of the metadata present at prep time. This is only tacked on if it's not actually present, since non-epipredict steps typically strip this.
These are then used in both `step_epi_ahead` and `step_epi_lag` if the signs are right to adjust the shift value.

### Added tests:
- arx_args_list: errors on `forecast_date`, `ahead`, `target_date` inconsistency
- arx_forecaster: 
  - warning if the `forecast_date` is ahead of the max time value
  - errors in the case of inconsistency
- `layer_add_forecast_date`
  - when `step_adjust_latency` is used previously, it adopts the date from that (specifically the `as_of` version)
- `layer_add_target_date`
  - when `step_adjust_latency` is used previously, it adopts the date from that (specifically the `as_of` version)
- snapshot tests
  - output arx_forecaster for extend_ahead is different and has the expected dates and output
  - output of arx_forecaster for locf
  - printing of arx_forecaster for various usecases
  - arx_forecaster manually setting the forecast date in the future works correctly with `adjust_latency`
- all sorts of tests for step_adjust_latency
  - the time adjustments are correct for extend_lags, extend_ahead
  - the results are equivalent to manually adjusting the lags/aheads
  - warnings if the ahead is too far in the future
  - locf correctly fills forward
  - multiple aheads are adjusted forward correctly
  - fixed_forecast_date and fixed_latency work
  - warning if there are relevant steps before `step_adjust_latency`
  - warns against `NA` removal early
  - adjusts lags/aheads for input with different lags per geo, compared with the by-hand version
  - specifying columns narrows the selection to only those columns
  - printing snapshots


@dajmcdon or @brookslogan anything you would want to see here?


I've realized that there are some fixes I need to do for the vignettes before we can merge this, but that shouldn't take too long

dsweber2:
ok, vignette fixed; there was a straggling `fill_locf`, and a previous problem where some versioned data wasn't available for a given `ref_time_value`.

R/utils-latency.R
```diff
@@ -0,0 +1,294 @@
...
+    arrange(across(all_of(key_colnames(x)))) %>%
+    ungroup() %>%
+    group_by(across(all_of(get_grouping_columns(x))))
+}
+
+#' return the names of the grouped columns, or `NULL`
+#' @param x an epi_df
+#' @keywords internal
+#' @importFrom utils head
+get_grouping_columns <- function(x) {
```
dajmcdon, line 219: This seems to be used in only one location. Where would the `groups` attr be set? Is this different from `key_colnames()`?

R/utils-latency.R
```diff
@@ -0,0 +1,294 @@
...
+  group_names <- names(attributes(x)$groups)
+  head(group_names, -1)
+}
+
+#' get the location of the last real value
+#' @param col the relevant column
+#' @param from_last instead of the number of columns
+#' @keywords internal
+count_single_column <- function(col, from_last) {
```
dajmcdon, line 228: Can this function just be `max(which(!is.na(col))`?

R/utils-latency.R
```diff
@@ -0,0 +1,294 @@
...
+  seq(from = from, to = to, by = by)
+}
+
+
+#' warn when the latency is larger than would be reasonable
+#' @param dataset the epi_df
+#' @param latency_table the whole collection of latencies
+#' @param target_columns the names of the columns that we're adjusting, and whether its unreasonably latent
+#' @keywords internal
+check_interminable_latency <- function(dataset, latency_table, target_columns, forecast_date, call = caller_env()) {
```
dajmcdon, line 255: If we're going to be producing these sorts of warnings, we should have a flag that turns it off.

R/utils-shift.R
```diff
@@ -0,0 +1,71 @@
...
+      shift_amount <- as.Date(as_of) - max_time
+      if (is.null(as_of) || is.na(as_of)) {
+        cli::cli_warn(paste(
+          "epi_data's `as_of` was {as_of}, setting to ",
+          "the latest time value, {max_time}."
+        ))
+        as_of <- max_time
+      } else if (as_of < max_time) {
+        cli::cli_abort(paste(
+          "`as_of` ({(as_of)}) is before the most ",
```
dajmcdon, line 37: Is there an extra set of parentheses around `as_of`?

R/utils-shift.R
```diff
@@ -0,0 +1,71 @@
...
+      }
+      effective_ahead <- as.integer(shift_amount + ahead)
+      time_type <- attributes(new_data)$metadata$time_type
+
+      if ((grepl("day", time_type) && (shift_amount >= 10)) ||
+        (grepl("week", time_type) && (shift_amount >= 4)) ||
+        ((time_type == "yearmonth") && (shift_amount >= 2)) ||
+        ((time_type == "yearquarter") && (shift_amount >= 1)) ||
+        ((time_type == "year") && (shift_amount >= 1))) {
+        cli::cli_warn(paste(
```
dajmcdon, line 50: Is this related to the `check_interminable_latency()` function above? We should have a way to turn this off.

vignettes/articles/symptom-surveys.Rmd
```diff
@@ -423,7 +423,7 @@ res_err4 <- res_all4 %>%
 knitr::kable(
   res_err4 %>%
     group_by(model, lead) %>%
-    summarize(err = median(err), n = length(unique(forecast_date))) %>%
+    summarise(err = median(err), n = length(unique(forecast_date))) %>%
```
dajmcdon, line 426: You don't like the Canadian spelling?

R/step_lag_difference.R
```diff
@@ -13,6 +13,7 @@
 #'
 #'
 #' @family row operation steps
+#' @importFrom cli cli_abort cli_warn
```
dajmcdon, line 16:
```suggestion

```

R/step_adjust_latency.R
```diff
@@ -0,0 +1,462 @@
...
+             "extend_ahead",
+             "locf",
+             "extend_lags"
+           ),
+           epi_keys_checked = c("geo_value"),
+           fixed_latency = NULL,
+           fixed_forecast_date = NULL,
+           skip = FALSE,
+           columns = NULL,
+           id = recipes::rand_id("adjust_latency")) {
```
dajmcdon, line 212: A few things here:
1. Unlike `{recipes}`, I don't think we should provide arguments that the user shouldn't touch. So `trained` and `columns` should be removed here and set explicitly to these values at the bottom of this function.
2. `role`. What is this used for? It's a bit ambiguous to me. You are using this step to potentially adjust outcomes or predictors depending on other arguments. So I think it should be removed unless I'm missing something.
3. `skip`. This is also potentially problematic. In standard usage, `FALSE` means that this step is evaluated at both train time and test time. I suspect that's what we want always. `TRUE` means it will NOT be evaluated at test time. This could be very strange, I think.
4. `rand_id` is imported.
5. are `...` intended to be used with tidy-selection? I think they should be empty, right? Unless `method = "locf"` maybe? In which case `role` may be needed?
6. I'm not sure about `epi_keys_checked`. What if the epi_df has other keys?

So my suggestion follows this model, but has many implications in the function body. And I'm not sure my understanding is correct.

```suggestion
  function(recipe,
                 method = c("extend_ahead", "locf", "extend_lags"),
           epi_keys_checked = c("geo_value"),
           fixed_latency = NULL,
           fixed_forecast_date = NULL,
           id = rand_id("adjust_latency")) {
```

R/step_adjust_latency.R
```diff
@@ -0,0 +1,462 @@
...
+    id = x$id
+  )
+}
+
+#' @importFrom dplyr %>% pull group_by_at
+#' @importFrom tidyr fill
+#' @export
+bake.step_adjust_latency <- function(object, new_data, ...) {
+  if (!isa(new_data, "epi_df")) {
+    # TODO if new_data actually has keys other than geo_value and time_value, this is going to cause problems
```
dajmcdon, line 371: we need to fix this before merging then.

R/step_adjust_latency.R
```diff
@@ -0,0 +1,462 @@
...
+    new_data <- as_epi_df(new_data)
+    attributes(new_data)$metadata <- object$metadata
+    attributes(new_data)$metadata$as_of <- object$forecast_date
+  }
+  if (object$method == "extend_ahead" || object$method == "extend_lags") {
+    attributes(new_data)$metadata$latency_method <- object$method
+    attributes(new_data)$metadata$shift_sign <- get_sign(object)
+    attributes(new_data)$metadata$latency_table <- object$latency_table
+    attributes(new_data)$metadata$forecast_date <- object$forecast_date
+    keys <- object$keys
```
dajmcdon, line 381: What is the reasoning for adding all of this as attributes? We should talk about it I think.

dajmcdon (changes requested):
I left lots of comments/suggestions. It may be easier to discuss some of these synchronously.

R/arx_forecaster.R
```diff
@@ -187,10 +228,25 @@ arx_fcast_epi_workflow <- function(
...
+#' @param forecast_date Date. The date from which the forecast is occurring.
+#'   The default `NULL` will determine this automatically from either
+#'   1. the maximum time value for which there's data if there is no latency
+#'   adjustment (the default case), or
+#'   2. the `as_of` date of `epi_data` if `adjust_latency` is
+#'   non-`NULL`.
+#' @param target_date Date. The date that is being forecast. The default `NULL`
+#'   will determine this automatically as `forecast_date + ahead`.
+#' @param adjust_latency Character or `NULL`. One of the `method`s of
+#'   [step_adjust_latency()], or `NULL` (in which case there is no adjustment).
```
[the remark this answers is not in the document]
dsweber2: yeah that's probably safer. I'll update it

R/epi_shift.R
```diff
@@ -2,43 +2,91 @@
...
+    dplyr::full_join,
+    by = ok
+  )
+  processed <- new_data %>%
+    full_join(shifted, by = ok) %>%
+    group_by(dplyr::across(dplyr::all_of(kill_time_value(ok)))) %>%
+    arrange(time_value) %>%
+    ungroup() %>%
+    as_epi_df()
+  attributes(processed)$metadata <- attributes(new_data)$metadata
```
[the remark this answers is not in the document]
dsweber2: This is to preserve metadata that was stored originally, such as `as_of`, extra `epi_keys`, and the new features such as `latency_table`. `as_epi_df` has no real way of knowing about those (e.g. the `as_of` ends up being today).

R/epi_recipe.R
```diff
@@ -65,15 +65,15 @@ epi_recipe.epi_df <-
   function(x, formula = NULL, ..., vars = NULL, roles = NULL) {
     if (!is.null(formula)) {
       if (!is.null(vars)) {
-        rlang::abort(
+        cli::cli_abort(
```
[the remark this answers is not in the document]
dsweber2: Styler's ok with that as a pattern, and I've generally used it too for cli stuff. Not sure why this one's different, I'll switch it over

R/flatline_forecaster.R
```diff
@@ -66,10 +66,7 @@ flatline_forecaster <- function(
   wf <- epi_workflow(r, eng, f)
   wf <- fit(wf, epi_data)
   preds <- suppressWarnings(forecast(
-    wf,
-    fill_locf = TRUE,
-    n_recent = args_list$nafill_buffer,
-    forecast_date = forecast_date
+    wf
```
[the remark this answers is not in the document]
dsweber2: I didn't really check the suppress warnings, if that was the original source it should be fine now

R/flatline_forecaster.R
```diff
@@ -108,6 +105,7 @@ flatline_forecaster <- function(
 #' flatline_args_list()
 #' flatline_args_list(symmetrize = FALSE)
 #' flatline_args_list(quantile_levels = c(.1, .3, .7, .9), n_training = 120)
+#' @importFrom cli cli_abort cli_warn
```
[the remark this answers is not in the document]
dsweber2: I'm never quite sure if the convention is to declare `@importFrom` at every instance of a function that uses it, or in some random function that does, or in some central location. I've generally erred towards everywhere, which is robust if redundant.

R/canned-epipred.R
```diff
@@ -103,10 +105,43 @@ print.canned_epipred <- function(x, name, ...) {
...
+
   cli::cli_ul(c(
     "{.val {n_geos}} unique geographic region{?s},",
     "At forecast date{?s}: {.val {fds}},",
-    "For target date{?s}: {.val {tds}}."
+    "For target date{?s}: {.val {tds}},"
   ))
+  fit_recipe <- extract_recipe(x$epi_workflow)
+  if (detect_step(fit_recipe, "adjust_latency")) {
+    is_adj_latency <- map_lgl(fit_recipe$steps, \(x) inherits(x, "step_adjust_latency"))
```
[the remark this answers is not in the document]
dsweber2: I find it a bit more readable, but yeah, not worth bumping just for this. I'll switch these to the older style

vignettes/articles/symptom-surveys.Rmd
dsweber2: truth be told, I use them 50/50, and couldn't tell you which one was :canada: XD

 I think this happened when I did a grep renaming because I was getting errors from my own use of `summarize`.

R/step_adjust_latency.R
dsweber2: It's really the only way to pass the values along. I suppose only one of `latency_method` and `shift_sign` is actually necessary, but the latency table is definitely needed, along with forecast_date.

I guess it might be more appropriate to store this in the recipe, but I'm not really sure about that (or how to go about it).

R/utils-latency.R
dsweber2: if `x` is already grouped, this means the step preserves any existing group structure, regardless of how it's grouped. 
```
a <- tibble(
+   geo_value = rep("place", 200),
+   time_value = as.Date("2021-01-01") + 0:199,
+   case_rate = sqrt(1:200) + atan(0.1 * 1:200) + sin(5 * 1:200) + 1,
+   death_rate = atan(0.1 * 1:200) + cos(5 * 1:200) + 1
+ ) %>% group_by(geo_value)
+ attributes(a)$groups
# A tibble: 1 × 2
  geo_value       .rows
  <chr>     <list<int>>
1 place           [200]
```
if `x` is ungrouped, this is just `null` so grouping by the result is a no-op.

Its a function not really for reuse but to reduce the overhead of reading `pad_to_end`, because the way of getting the previous grouping is a bit jank.

R/utils-latency.R
dsweber2: yep, that is definitely the right way to do this

R/utils-shift.R
dsweber2: This... is actually the older implementation, and should be removed entirely

R/step_adjust_latency.R
dsweber2: Most of that tracks, though 
5. `...` are for tidy selection, if for whatever reason you don't want to apply it to every column, or if you want to apply it to specifically non-raw columns.
6. `epi_keys_checked` should probably default to all epi_keys besides time, so I'll switch it to that.

R/step_adjust_latency.R
dsweber2: That was actually an old note that I fixed and forgot to drop. The fix is somewhat suboptimal, because it drags the metadata from the prep step, rather than the `new_data` at the very start of the recipe.

R/arx_classifier.R
```diff
@@ -171,11 +190,25 @@ arx_class_epi_workflow <- function(
...
+  }
   r <- r %>%
-    step_epi_ahead(!!o, ahead = args_list$ahead, role = "pre-outcome") %>%
-    recipes::step_mutate(
-      outcome_class = cut(!!o2, breaks = args_list$breaks),
+    step_epi_ahead(!!pre_out_name, ahead = args_list$ahead, role = "pre-outcome")
+  r <- r %>%
+    step_mutate(
+      across(
+        starts_with("ahead"),
```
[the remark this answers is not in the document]
dsweber2: I initially tried to do this using `all_outcomes()`, but that doesn't seem to work for some reason. I'm ok saying that the user shouldn't be starting column names with `"ahead_"`, given that its clearly a key phrase

R/flatline_forecaster.R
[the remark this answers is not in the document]
dsweber2: so there are still warnings apparently: 
```
Warning message:
In formals(fun) : argument is not a function
```
I'm struggling to get it to produce a line number or anything, so I'm just going to leave the suppressWarnings in

R/utils-latency.R
dsweber2: should also be in the canned forecaster

R/utils-latency.R
```diff
@@ -0,0 +1,294 @@
...
+    if (inherits(recipe_step, rel_step_type)) {
+      recipe_columns <- recipes_eval_select(recipe_step$terms, recipe$template, recipe$term_info)
+      if (any(recipe_columns %in% terms_used)) {
+        return(list(term = recipe_columns, shift = recipe_step[shift_name], prefix = recipe_step$prefix))
+      }
+    }
+    return(NULL)
+  }
+  rel_list <- recipe$steps %>%
+    purrr::map(extract_named_rates) %>%
```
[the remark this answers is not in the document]
dsweber2: Whatever happened between when I first ran this and when I just tried removing this, it seems to have started working. I guess we can drop the `purrr` dependency, but I would much rather not have these shims around as a potential source of bugs/maintenance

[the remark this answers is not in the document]
dajmcdon: FWIW, these shims are read-only from `rlang`. So we've still outsourced the bugs+maintenance...

There's some background at r-lib/usethis#1654

These are imported/updated with `usethis::use_standalone(repo, name_of_file)`, e.g. `usethis::use_standalone("r-lib/rlang", "purrr")`.

R/canned-epipred.R
```diff
@@ -103,10 +105,43 @@ print.canned_epipred <- function(x, name, ...) {
...
+      type_str <- "Lags"
+    } else {
+      step_names <- ""
+      type_str <- "columns locf"
+    }
+    later_steps[[1]]$columns
+    valid_columns <- later_steps %>%
+      keep(\(x) inherits(x, step_names)) %>%
+      purrr::map("columns") %>%
+      reduce(c)
```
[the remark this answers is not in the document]
dsweber2: right, this is the one that the patch doesn't support, since `"columns"` isn't a function.

DESCRIPTION
```diff
@@ -38,6 +38,8 @@ Imports:
     magrittr,
     recipes (>= 1.0.4),
     rlang (>= 1.0.0),
+    purrr,
+    smoothqr,
```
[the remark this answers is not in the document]
dsweber2: I suppose not; I added it before realizing that the problem with pkgdown was from the `_pkgdown.yml` file

R/canned-epipred.R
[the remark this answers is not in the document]
dajmcdon: Sorry, the suggestion should have also included line 136 below. There `valid_columns` needs to be `valid_col_names`. That has no errors.

R/utils-latency.R
```diff
@@ -0,0 +1,294 @@
...
+#' @importFrom tidyselect all_of
+#' @importFrom rlang list2
+#' @importFrom vctrs vec_cast
+#' @importFrom dplyr across arrange bind_rows group_by summarise
+#' @keywords internal
+pad_to_end <- function(x, groups, end_date, columns_to_complete = NULL) {
+  if (is.null(columns_to_complete)) {
+    columns_to_complete <- setdiff(names(x), key_colnames(x))
+  }
+  itval <- epiprocess:::guess_period(c(x$time_value, end_date), "time_value")
```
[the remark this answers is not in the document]
brookslogan: Would consulting `time_type `/ `guess_time_type` work / be more appropriate here?  `guess_period` can output periods that aren't supported by e.g. `epi_slide_opt`

dsweber2:
Ok, a minor amount of changes necessary to that PR to get it working w/out modifying the way prep.epi_recipe works. Are we  good to merge?

dajmcdon:
It’s still pretty big, so give me a bit of time to take another pass. Sent from my iPhoneOn Sep 20, 2024, at 10:13, David Weber ***@***.***> wrote:﻿
Ok, a minor amount of changes necessary to that PR to get it working w/out modifying the way prep.epi_recipe works. Are we  good to merge?

R/arx_classifier.R
```diff
@@ -171,11 +190,25 @@ arx_class_epi_workflow <- function(
         )
     }
   }
-  o2 <- rlang::sym(paste0("ahead_", args_list$ahead, "_", o))
+  ahead_out_name <- glue::glue("ahead_[0-9]*_{pre_out_name}")
```
dajmcdon, line 193: Can you explain what this should do? I see:
``` r
pre_out_name <- "pre_out_name" # just to examine
glue::glue("ahead_[0-9]*_{pre_out_name}")
#> ahead_[0-9]*_pre_out_name
```

R/arx_classifier.R
```diff
@@ -171,11 +190,25 @@ arx_class_epi_workflow <- function(
...
+  method_adjust_latency <- args_list$adjust_latency
+  if (method_adjust_latency != "none") {
+    # only extend_ahead is supported atm
```
dajmcdon, line 196: Stale?
```suggestion
```

R/arx_classifier.R
```diff
@@ -297,10 +327,13 @@ arx_class_args_list <- function(
...
-        "`forecast_date` + `ahead` must equal `target_date`.",
-        i = "{.val {forecast_date}} + {.val {ahead}} != {.val {target_date}}."
-      ))
+      cli_warn(
+        c(
+          "`forecast_date` + `ahead` must equal `target_date`.",
+          i = "{.val {forecast_date}} + {.val {ahead}} != {.val {target_date}}."
+        ),
+        class = "epipredict__arx_args__inconsistent_target_ahead_forecaste_date"
+      )
```
dajmcdon, line 336: Similar comment in the forecaster below.

```suggestion
      cli_warn("`forecast_date`  {.val {forecast_date}} + `ahead` {.val {ahead}} must equal `target_date` {.val {target_date}}.",
        class = "epipredict__arx_args__inconsistent_target_ahead_forecaste_date"
      )
```

R/arx_forecaster.R
```diff
@@ -51,13 +51,17 @@ arx_forecaster <- function(
...
-    fill_locf = TRUE,
-    n_recent = args_list$nafill_buffer,
-    forecast_date = args_list$forecast_date %||% max(epi_data$time_value)
-  ) %>%
-    as_tibble() %>%
+  # get the forecast date for the forecast function
+  if (args_list$adjust_latency == "none") {
+    forecast_date_default <- max(epi_data$time_value)
+  } else {
+    forecast_date_default <- attributes(epi_data)$metadata$as_of
```
dajmcdon, line 58: This is about the 5th time I've seen this line. Can we make a utility:
```r
get_edf_asof <- function(x) attr(x, "metadata)$as_of
```

R/arx_forecaster.R
```diff
@@ -51,13 +51,17 @@ arx_forecaster <- function(
...
+  }
+  forecast_date <- args_list$forecast_date %||% forecast_date_default
+
+
+  preds <- forecast(wf, forecast_date = forecast_date) %>%
+    tibble::as_tibble() %>%
```
dajmcdon, line 64:
```suggestion

  preds <- forecast(wf, forecast_date = forecast_date) %>%
    as_tibble() %>%
```

R/arx_forecaster.R
```diff
@@ -119,22 +123,62 @@ arx_fcast_epi_workflow <- function(
...
+  target_date <- args_list$target_date %||% (forecast_date + args_list$ahead)
+  if (forecast_date + args_list$ahead != target_date) {
+    cli_abort(
+      c(
+        "`forecast_date` + `ahead` must equal `target_date`.",
+        i = "{.val {forecast_date}} + {.val {ahead}} != {.val {target_date}}."
+      ),
+      class = "epipredict__arx_forecaster__inconsistent_target_ahead_forecaste_date"
+    )
+  }
```
dajmcdon, line 150: Something like this is my preference if it passes styler. The `c(` on its own line kinda bugs me.

```suggestion
    cli_abort(
        "`forecast_date` {.val {forecast_date}} + `ahead` {.val {ahead}} must equal `target_date` {.val {target_date}}.",
      class = "epipredict__arx_forecaster__inconsistent_target_ahead_forecaste_date"
    )
  }
```

R/arx_forecaster.R
```diff
@@ -258,16 +309,18 @@ arx_args_list <- function(
...
-        "`forecast_date` + `ahead` must equal `target_date`.",
-        i = "{.val {forecast_date}} + {.val {ahead}} != {.val {target_date}}."
-      ))
+      cli_abort(
+        c(
+          "`forecast_date` + `ahead` must equal `target_date`.",
+          i = "{.val {forecast_date}} + {.val {ahead}} != {.val {target_date}}."
+        ),
+        class = "epipredict__arx_args__inconsistent_target_ahead_forecaste_date"
+      )
```
dajmcdon, line 323: Same again.

R/epi_recipe.R
dajmcdon, line 1: Can we revert this entire file to match current `dev`? Or is there something new that you need here?

R/utils-enframer.R
dajmcdon, line 1: The `enframer()` function is used nowhere afaict. Do you mind deleting it and the associated tests?

dajmcdon (approved):
A few remaining minor points, but otherwise looking good!

R/arx_classifier.R
dsweber2: probably? I've added an issue https://github.com/cmu-delphi/epipredict/issues/389 to explicitly add tests

R/arx_classifier.R
dsweber2: oh this is a regex that catches anything that looks like `ahead_<any set of numbers>_<pre_out_name>`

R/arx_forecaster.R
dsweber2: I definitely think this is a good idea, but should probably be an issue over in `epiprocess`

R/epi_recipe.R
dsweber2: changes here:
- various cli_aborts
- some unneeded `::` being removed
-  a use of `c` for a warning when `paste0` was intended
- `i` -> `ii` to improve searchability
we could drop all this, but I see no reason we should

R/utils-enframer.R
dsweber2: sure

R/epi_recipe.R
dajmcdon: I'm just anticipating potential merge challenges with another PR. Up to you.

dsweber2: if you'd rather add them manually rather than have a merge conflict resolution sort it out I can remove

R/arx_classifier.R
dajmcdon: Is it missing some data to key on? unless I'm missing something, it just prints the regex, not the result of the regex.

R/arx_classifier.R
dajmcdon: Then other methods should error?

dsweber2: I think they probably work; is resolving whether they do or not blocking?

dajmcdon: As a reviewer, I see a comment here that doesn't match the surrounding code. If they work, then let's remove the comment. If we don't know if they work, or don't think they work, then let's put an error here that catches if the user asks for them and refuses to procede.

dsweber2: got it, comment removed and moved to an issue

R/arx_classifier.R
dsweber2: `ahead_out_name` is used in a `matches`, which uses regex. Its in a separate variable to avoid execution shenanigans. renaming to make it clear it should just be a regex

R/arx_classifier.R
dsweber2: Alright, explicitly tested it and added errors b/c it wasn't working. Then again, the user would've hit the errors I got anyway, so this is just adding a more explicit wrapper patch.

R/arx_classifier.R
dajmcdon: I understand how it gets used. My concern is: 
```r
 glue::glue("ahead_[0-9]*_{pre_out_name}")
```

Why is `[0-9]*` hardcoded into the name? Glue doesn't reverse regex it. Did you mean this to be `{ahead}`? 
Whereas (if regex matching) later on you can use `[0-9]*` to match the result (in the `ahead_out_name` object)?

dsweber2: `ahead` may not be the actual name of the column, because that value could change b/c of ahead adjustment. I'm not expecting glue to do anything with `[0-9]*` but pass it on, since `ahead_out_name` is a regex (which are just strings). `ahead_[0-9]*_pre_out_name` is the expected result for `ahead_out_name_regex`.

PR #310, opened 2024-04-04
feat: canned forecasters print their args by default

### LINKED ISSUE #309: Print the arguments used by a canned forecaster by default
There are many hidden arguments/parameters set by default in a canned forecaster. We should print these by default and add a `.verbose` option to suppress this print.

dshemetov:
### Checklist

Please:

-   [x] Request a review from one of the current epipredict main reviewers:
        dajmcdon.
-   [ ] ~~Makes sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
        version number (the second number).~~

Adds `.verbose = TRUE` to canned forecasters and prints their arguments if TRUE.

-   Resolves #309

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@ Pre-1.0.0 numbering scheme: 0.x will indicate releases, while 0.0.x will indicat
 
 # epipredict 0.1
 
+- canned forecaster workflow functions (such as `arx_forecaster()`) now display their parameters by default; this can be turned off with `.verbose = FALSE`
 - `*_args_list()` functions now warn if `forecast_date + ahead != target_date`
 - `layer_residual_quantiles()` will now error if any of the residual quantiles are NA
 - add `check_enough_train_data` that will error if training data is too small
--- a/R/arx_classifier.R
+++ b/R/arx_classifier.R
@@ -18,6 +18,7 @@
 #'   also be used.
 #' @param args_list A list of customization arguments to determine
 #'   the type of forecasting model. See [arx_class_args_list()].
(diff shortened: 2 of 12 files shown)


dajmcdon (commented):
Two comments:
1. I'd prefer this to be FALSE by default. I don't think functions should print stuff unless asked. It's potentially useful for debugging purposes, but otherwise the documentation (and, in this case, the function defaults) should be clear enough.
2. Should it be in the workflow creator? Or just in the forecaster? I can see arguments both ways. Another option is to set the default to TRUE in the workflow creator (a more developer-facing function) while it's FALSE in the forecaster (typical user-facing).

dshemetov:
1. I think I leaned to the noisy default because I figured this is most useful for a beginner user of this package and I didn't want them to have to hunt for this flag. The more experienced users can turn it off and write a default-off wrapper for themselves if the extra typing is really bothering them.
2. My thinking was that both functions are user-facing, the workflow creator is just for a more advanced user. If that's true, I could see defaulting to off in the workflow creator and on in the simpler one.

dajmcdon:
In R, I don't think it's customary to print things by default (with the possible exception of a progress bar, though even that is usually silent). I'd prefer these to be off. 

Come to think of it, I think we should add these to the print methods for the resulting object rather than altering the function signature at all.

dshemetov:
I hear ya about not wanting to be too noisy. 

I like the idea of adding this to the object printer. Let me see what I can do there.

dshemetov:
After thinking about it some more, doesn't seem worth it.

PR #314, opened 2024-04-08
Smooth qr vignette

rachlobay:
This branch has the smooth qr vignette. I removed the lower and upper miscoverage rate stuff.


DESCRIPTION
```diff
@@ -36,6 +36,7 @@ Imports:
     glue,
     hardhat (>= 1.3.0),
     magrittr,
+    purrr,
```
dajmcdon, line 39: in Suggests

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
+---
+title: "Smooth Quantile Regression"
+---
+
```
dajmcdon, line 4: Same comment as #313

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+```{r setup, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = FALSE,
+  comment = "#>",
+  warning = FALSE,
+  message = FALSE,
+  cache = TRUE
```
dajmcdon, line 11:
```suggestion
  out.width = "100%"
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+)
+```
+
+# Introducing smooth quantile regression
+
+Whereas the standard application of time-series forecasting techniques has been to forecast single horizons, in multi-period forecasting, the goal is to forecast several horizons simultaneously. The standard application of these techniques aims to predict the signal for a single forecast horizon (or "ahead"), most often one-step-ahead. This is useful in epidemiological applications where decisions are often based on the trend of a signal.
```
dajmcdon, line 17:
```suggestion
Whereas other time-series forecasting examples in this package have used (direct) models for single horizons, in multi-period forecasting, the goal is to (directly) forecast several horizons simultaneously. This is useful in epidemiological applications where decisions are often based on the trend of a signal.
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+The idea underlying smooth quantile regression that the future can be approximated by a smooth curve. So this approach enforces smoothness across the horizons for point estimation by regression or interval prediction by quantile regression. Our focus in this chapter is the latter.
```
dajmcdon, line 19:
```suggestion
The idea underlying smooth quantile regression that set forecast targets can be approximated by a smooth curve. So this approach enforces smoothness across the horizons for point estimation by regression or interval prediction by quantile regression. Our focus in this chapter is the latter.
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+  engine = "smoothqr",
+  outcome_locations = NULL,
+  quantile_levels = 0.5,
+  degree = 3L
+)
+```
+
+For smooth quantile regression, the type of model or `mode` is regression.
+
+The only `engine` that is currently supported is `smooth_qr()` from the [`smooth_qr` package](https://dajmcdon.github.io/smoothqr/).
```
dajmcdon, line 37:
```suggestion
The only `engine` that is currently supported is `smooth_qr()` from the [`smoothqr` package](https://dajmcdon.github.io/smoothqr/).
```

vignettes/smooth-qr.Rmd
dajmcdon: Also, we should cite the paper early on: https://arxiv.org/abs/2202.09723

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+The `outcome_locations` indicate the multiple horizon (ie. ahead) values. These should be specified by the user.
+
+The `quantile_levels` parameter is a vector of values that indicates the quantiles to be estimated. The default is the median (0.5 quantile).
+
+The `degree` parameter indicates the number of polynomials used for smoothing of the response. It should be no more than the number of responses. If the degree is precisely equal to the number of aheads, then there is no smoothing. To better understand this parameter and how it works, we should look to its origins and how it is used in the model.
```
dajmcdon, line 43:
```suggestion
The `degree` parameter indicates the degree of the polynomials used for smoothing of the response. It should be no more than the number of aheads. If the degree is precisely equal to the number of aheads, then there is no smoothing. To better understand this parameter and how it works, we should look to its origins and how it is used in the model.  
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+
+The `outcome_locations` indicate the multiple horizon (ie. ahead) values. These should be specified by the user.
+
+The `quantile_levels` parameter is a vector of values that indicates the quantiles to be estimated. The default is the median (0.5 quantile).
+
+The `degree` parameter indicates the number of polynomials used for smoothing of the response. It should be no more than the number of responses. If the degree is precisely equal to the number of aheads, then there is no smoothing. To better understand this parameter and how it works, we should look to its origins and how it is used in the model.  
+
+# Model form 
+
+Smooth quantile regression is linear auto-regressive, with the key feature being a transformation that forces the coefficients to satisfy a constraint. The purpose if this is for each model coefficient to be a smooth function of ahead values, and so each such coefficient is set to be a linear combination of smooth basis functions (such as a spline or a polynomial).
```
dajmcdon, line 47: This is not the easiest interpretation. It's actually the fitted values that you want to satisfy the constraint $X\hat B$ not $\hat B$. But because things are linear, this implies a constraint $\hat B$.

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+```
+
+And for plotting our focus will be on a subset of two states - California and Utah.
+
+```{r}
+geos <- c("ut", "ca")
+```
+
+Suppose that our goal with this data is to predict COVID-19 death rates at several horizons for each state. On day $t$, we want to predict new deaths $y$ that are $a = 1,\dots, 28$ days ahead at locations $j$ using the death rates from today, 1 week ago, and 2 weeks ago. So for each location, we'll predict the median (0.5 quantile) for each of the target dates by using
+$$
```
dajmcdon, line 87: Maybe use $b$ not $\beta$, to match the paper.

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+smooth_preds <- smooth_fc(edf, fd = fd)
+
+head(smooth_preds)
+```
+Most often, we're not going to want to limit ourselves to just predicting the median value as there is uncertainty about the predictions, so let's try to predict several different quantiles in addition to the median:
+
+```{r, warning = FALSE}
+smooth_preds <- smooth_fc(edf, quantiles = c(.1, .25, .5, .75, .9), fd = fd)
+
+head(smooth_preds)
```
dajmcdon, line 160:
```suggestion
smooth_preds
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,305 @@
...
+```
+
+Notice that we allow the function user to specify the aheads, degree, and quantile as they may want to change these parameter values. We also allow for input of the forecast date as we fixed that at the onset of this demonstration.
+
+We now can produce smooth quantile regression predictions for our problem:
+
+```{r, warning = FALSE}
+smooth_preds <- smooth_fc(edf, fd = fd)
+
+head(smooth_preds)
```
dajmcdon, line 153:
```suggestion
smooth_preds
```

dajmcdon (changes requested):
I looked at the top part. This is where you were going to use WIS right?

rachlobay:
Yes. This is the vignette where I am going to try add a bit about WIS to. TBD.

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+One way to quantify the impact of these on the forecasting is to look at the mean absolute error (MAE) or mean squared error (MSE) over the degrees. We can select the degree that results in the lowest MAE. 
+
+Since the MAE compares the predicted values to the actual values, we will first join the test data to the predicted data for our comparisons:
+```{r, message = FALSE}
+tedf_sub <- tedf %>%
+  rename(target_date = time_value, actual = death_rate) %>%
+  select(geo_value, target_date, actual)
+
+smooth_preds_df_deg <- smooth_preds_list %>%
+  left_join(tedf_sub)
```
dajmcdon, line 223:
```suggestion
  left_join(tedf_sub, by = c("geo_value", "target_date"))
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+```
+
+And then compute the MAE for each of the degrees:
+```{r, message = FALSE}
+smooth_preds_df_deg <- smooth_preds_list %>%
+  left_join(tedf_sub) %>%
```
dajmcdon, line 229: This was already done in the previous chunk, right?

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+plot_preds(baseline_preds, geos, bind_rows(tedf, edf), fd)
+```
+
+Unlike for smooth quantile regression, the resulting forecasts are not smooth curves, but rather jagged and irregular in shape.
+
+For a more formal comparison between the two approaches, we could compare the test performance in terms of accuracy through calculating either the, MAE or MSE, where the performance measure of choice can be calculated over over all times and locations for each ahead value
+
+```{r, message = FALSE}
+baseline_preds_mae_df <- baseline_preds %>%
+  left_join(tedf_sub) %>%
```
dajmcdon, line 270:
```suggestion
  left_join(tedf_sub, by = c("geo_value", "target_date"))
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+  group_by(ahead) %>%
+  mutate(error = abs(`0.5` - actual)) %>%
+  summarise(mean = mean(error)) %>%
+  mutate(type = "baseline")
+
+smooth_preds_mae_df <- smooth_preds %>%
+  left_join(tedf_sub) %>%
```
dajmcdon, line 277:
```suggestion
  left_join(tedf_sub, by = c("geo_value", "target_date"))
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+  mutate(error = abs(`0.5` - actual)) %>%
+  summarise(mean = mean(error)) %>%
+  mutate(type = "smooth")
+
+preds_mae_df <- bind_rows(baseline_preds_mae_df, smooth_preds_mae_df)
+
+ggplot(preds_mae_df, aes(ahead, mean, color = type)) +
+  geom_line() +
+  xlab("Ahead") +
+  ylab("Mean MAE")
```
dajmcdon, line 288: Let's not use these colours.

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+or over all aheads, times, and locations for a single numerical summary.
+
+```{r}
+mean(baseline_preds_mae_df$mean)
+mean(smooth_preds_mae_df$mean)
+```
+
+The former shows that forecasts for the immediate future and for the distant future are more inaccurate for both models under consideration. The latter shows that the smooth quantile regression model and baseline models perform very similarly overall, with the smooth quantile regression model only slightly beating the baseline model in terms of overall average MAE.
+
+One other commonly used metric is the WIS score ([ Bracher et al., 2021](https://arxiv.org/pdf/2005.12881.pdf)), which a scoring rule that is based on the population quantiles. The point is to score the interval, whereas MAE only evaluates the accuracy of the point forecast.
```
dajmcdon, line 300:
```suggestion
One other commonly used metric is the Weighted Interval Score (WIS, [Bracher et al., 2021](https://arxiv.org/pdf/2005.12881.pdf)), which a scoring rule that is based on the population quantiles. The point is to score the interval, whereas MAE only evaluates the accuracy of the point forecast.
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+Suppose we have $F$ be a forecaster composed of predicted quantiles $q_{\tau}$ for the set of quantile levels $\tau$. Then, in terms of the predicted quantiles, the WIS for target variable $Y$ is represented as follows ([McDonald et al., 2021](https://www.pnas.org/doi/full/10.1073/pnas.2111453118)):
```
dajmcdon, line 302:
```suggestion
Let $F$ be a forecast composed of predicted quantiles $q_{\tau}$ for the set of quantile levels $\tau$. Then, in terms of the predicted quantiles, the WIS for target variable $Y$ is represented as follows ([McDonald et al., 2021](https://www.pnas.org/doi/full/10.1073/pnas.2111453118)):
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+    na.rm = TRUE
+  ))
+}
+```
+
+Next, we apply the `wis_dist_quantile` function to get a WIS score for each state on each target date. We then compute the mean WIS for each ahead value over all of the states. The results for each of the smooth and baseline forecasters are shown in a similar style line plot as we chose for MAE:
+
+```{r}
+smooth_preds_wis_df <- smooth_preds %>%
+  left_join(tedf_sub) %>%
```
dajmcdon, line 334:
```suggestion
  left_join(tedf_sub, by = c("geo_value", "target_date"))
```

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+  group_by(ahead) %>%
+  summarise(mean = mean(wis)) %>%
+  mutate(type = "baseline")
+
+preds_wis_df <- bind_rows(smooth_preds_wis_df, baseline_preds_wis_df)
+
+ggplot(preds_wis_df, aes(ahead, mean, color = type)) +
+  geom_line() +
+  xlab("Ahead") +
+  ylab("Mean WIS")
```
dajmcdon, line 354: Better colours.

dajmcdon (approved):
Let's move this into `articles/` since it takes a _long_ while to run.

The rest of my comments are minor. Looks good!!

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+And then compute the MAE for each of the degrees:
+```{r, message = FALSE}
+smooth_preds_df_deg <- smooth_preds_list %>%
+  left_join(tedf_sub) %>%
+  group_by(degree) %>%
+  mutate(error = abs(`0.5` - actual)) %>%
+  summarise(mean = mean(error))
+
+# Arrange the MAE from smallest to largest
+head(smooth_preds_df_deg %>% arrange(mean))
```
dajmcdon, line 235:
```suggestion
smooth_preds_df_deg %>% arrange(mean)
```

dajmcdon (commented):
Missed one.

vignettes/smooth-qr.Rmd
```diff
@@ -0,0 +1,374 @@
...
+plot_preds(smooth_preds, geos, bind_rows(tedf, edf), fd)
+```
+
+We can see that the predictions are smooth curves for each state, as expected when using smooth quantile regression. In addition while the curvature of the forecasts matches that of the truth, the forecasts do not look remarkably accurate.
+
+## Varying the degrees parameter
+
+We can test the impact of different degrees by using the `map()` function. Let's try out all degrees from 1 to 7:
+
+```{r, warning = FALSE}
```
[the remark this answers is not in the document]
rachlobay: Noted

vignettes/smooth-qr.Rmd
rachlobay: Added a citation near to the beginning of the vignette

vignettes/smooth-qr.Rmd
rachlobay: Yes. Removed one of the redundancies.

vignettes/smooth-qr.Rmd
rachlobay: For now, will leave as is.

PR #328, opened 2024-04-30
325 workflow adj

### LINKED ISSUE #325: Update the add / remove / update functions to make sure they do not strip the `epi_workflow` class.
From a discussion with Daniel: `workflows::add_model()` is a function, not an S3 generic (like `fit()`). But we created it in our package as a generic. Then we made methods for it like `add_model.epi_workflow()`. But within those, we called things like `workflows::add_model()`. This strips the `epi_workflow` class from the object.
I’m not sure how we should create our versions. If someone loads workflows after they load epipredict, it will overwrite our generic + methods. This is why we had to move the `update.Rmd` file into the articles folder (otherwise R CMD check is finding an error based on this where the `update_model()` function ~ line 170 is stripping the `epi_workflow` class). We should update the vignette too and can move it back to the vignettes folder after this issue is handled.

The “right” think to do is probably to do a pull request into workflows. But for now, we may have to replace all the add / remove / update functions with add_epi / remove_epi / update_epi versions.

dajmcdon:
### Checklist

Please:

- [x] Request a review from one of the current epipredict main reviewers.

- Resolves #325


rachlobay (approved):
Looks good to me!

PR #331, opened 2024-05-09
Djm/deprecate

dajmcdon:
### Checklist

Please:

      dajmcdon.

In reviewing #296 , some of the error message changes I was suggesting seemed to borrow from existing tidymodels functions. All of these are internal, and we don't actually need them.

@dsweber2 note also the internal versions of purrr functions. This avoids the need to import the package. These are "in" `{rlang}` and accessible as `usethis::use_standalone("r-lib/rlang", "purrr")`, for example.


R/compat-purrr.R
dsweber2, line 1: Why are we maintaining our own versions of `purrr` functions? This doesn't make any sense.

dajmcdon: We are using the `rlang` shim. Before, we were using it by me copying from the GitHub repo and pasting into the file in `epipredict`. Now, we import it directly, fully. But it's missing 3 functions that we use. So those are here.

dsweber2: I see. I still am not clear on why we're not just using the real versions, but I'm not going to stop things moving forward because of it.

dajmcdon: Typically, I prefer to avoid writing new functions which are written elsewhere. But depending on upstream packages has many potential downsides. 
1. For example, `{purrr}` deprecated `map_dfr()` which we use. There's no assurance that they don't make other changes in the future. 
2. CRAN checks complain when you have >20 Imports.
3. We only use these functions internally, so best to avoid the overhead. 

I'd also provide a caveat to the general programming recommendation (avoid writing new functions...). 

> Do whatever Hadley Wickham does.

Note that the same `{purrr}` shim is used in [`{dplyr}`](https://github.com/tidyverse/dplyr/blob/d69802224a1df16d7a795ce313880116ea62ed6e/R/import-standalone-purrr.R) as well as many other packages.

tests/testthat/test-check_enough_train_data.R
```diff
@@ -74,7 +74,6 @@ test_that("check_enough_train_data outputs the correct recipe values", {
 
   expect_equal(nrow(p), 2 * n)
   expect_equal(ncol(p), 4L)
-  expect_s3_class(p, "epi_df")
```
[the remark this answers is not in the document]
dsweber2: Same question for all of the below

dsweber2 (commented):
I'm a fan of dropping unnecessary complexity, but I don't understand `bake.epi_recipe` and `epi_juice` well enough to be sure we can drop them. Did you run any additional tests that didn't get formalized, and/or which tests do you expect would have thrown a problem if we actually needed `bake.epi_recipe` and `epi_juice`?

tests/testthat/test-check_enough_train_data.R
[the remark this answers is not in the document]
dajmcdon: Good catch. I was thinking of the predictions not the training data. I'll have to make some fixes.

NEWS.md
```diff
@@ -43,3 +43,6 @@ Pre-1.0.0 numbering scheme: 0.x will indicate releases, while 0.0.x will indicat
 - `arx_fcast_epi_workflow()` and `arx_class_epi_workflow()` now default to
   `trainer = parsnip::logistic_reg()` to match their more canned versions.
 - add a `forecast()` method simplify generating forecasts
+- remove unnecessary `bake.epi_recipe()` and `epi_juice()`.
```
dajmcdon, line 46:
```suggestion
- refactor `bake.epi_recipe()` and remove `epi_juice()`. 
```

dajmcdon:
@dsweber2 OK. I refactored `bake.epi_recipe()` to fix the issue with the tests I'd removed. This now results (correctly) in an `epi_df`. But note that the column order is different. It is now the order enforced by `epiprocess::as_epi_df`.

Also added a test to check that my refactor works as expected.

`epi_juice()` was internal to `bake.epi_recipe()` and is no longer needed for anything.

dsweber2 (approved):
lgtm. one weird thing I'm getting locally is `man/add_model.Rd` is getting renamed to `man/Add_model.Rd` when I run `document`

dajmcdon:
> ...`man/add_model.Rd` is getting renamed to `man/Add_model.Rd` when I run `document`

Mine as well. I think it's like that on `dev`.

PR #334, opened 2024-05-15
initial layer adjustments

dsweber2:
This is to keep the discussion on the layer additions a little bit separate from the other PR, because that is becoming unwieldy in length. The documentation is still a WIP.


dsweber2:
ok, also added in some stuff to deal with inter-geo latency, based on @lcbrooks discussion on the other PR. The option is called `epi_keys_checked`, and defaults to just `"geo_value"`, but can group by any of the `epi_keys`

R/step_adjust_latency.R
```diff
@@ -13,30 +13,42 @@
...
-#'   the amount to offset the ahead or lag by. If a single integer, this is used
-#'   for all columns; if a labeled vector, the labels must correspond to the
-#'   base column names.  If `NULL`, the latency is the distance between the
-#'   `epi_df`'s `max_time_value` and either the `fixed_asof` or the `epi_df`'s
-#'   `as_of` field.
-#' @param fixed_asof either a date of the same kind used in the `epi_df`, or
-#'   NULL. Cannot be set at the same time as `fixed_latency`. If a date, it
-#'   gives the date from which the forecast is actually occurring. If `NULL`,
-#'   the `as_of` is determined either from `fixed_latency` or automatically.
+#'   vector. Cannot be set at the same time as `fixed_asof`. If non-`NULL`, the
```
dajmcdon, line 40: wasn't `fixed_asof` removed?

dsweber2: renamed to `fixed_forecast_date`, good catch thanks

R/utils-latency.R
```diff
@@ -135,52 +136,98 @@ set_asof <- function(new_data, info) {
...
+    # get the as_of in an `adjust_latency` step, regardless of where
+    handpicked_as_of <- map(
+      this_recipe$steps,
+      function(x) {
+        if (inherits(this_recipe$steps[[3]], "step_adjust_latency")) x$as_of
+      }
+    ) %>% Filter(Negate(is.null), .)
+    if (length(handpicked_as_of) > 0) {
+      max_time_value <- handpicked_as_of[[1]]
+    } else {
```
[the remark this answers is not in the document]
dsweber2: the `this_recipe$steps[[3]]` was left over from development, surprised it didn't cause any errors yet.

`handpicked_as_of` _should_ have only value (otherwise they have multiple `step_adjust_latency`s, which shouldn't work). I suppose I should add a check that there's only one `step_adjust_latency` during step creation

R/utils-latency.R
```diff
@@ -135,52 +136,98 @@ set_asof <- function(new_data, info) {
...
-  }
-  as_of <- attributes(new_data)$metadata$as_of
-  max_time <- max(time_values)
+    {
+      # null and "" don't work in `group_by`
+      if (!is.null(epi_keys_checked) && epi_keys_checked != "") {
+        group_by(., get(epi_keys_checked))
+      } else {
+        .
+      }
```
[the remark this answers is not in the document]
dsweber2: sure, I was just following the convention in the function previously of "in-piping" the logic.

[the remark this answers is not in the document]
dsweber2: and by in the function previously, I actually mean in `arx_forecaster` https://github.com/cmu-delphi/epipredict/blob/de6e1db8919aadd44b639da2a0f8860a6af1d358/R/arx_forecaster.R#L129-L146

R/step_adjust_latency.R
```diff
@@ -13,30 +13,42 @@
...
+#'   amount to offset the ahead or lag by. If a single integer, this is used for
+#'   all columns; if a labeled vector, the labels must correspond to the base
+#'   column names (before lags/aheads).  If `NULL`, the latency is the distance
+#'   between the `epi_df`'s `max_time_value` and either the
+#'   `fixed_forecast_date` or the `epi_df`'s `as_of` field (the default for
+#'   `forecast_date`).
+#' @param fixed_forecast_date either a date of the same kind used in the
+#'   `epi_df`, or `NULL`. Exclusive with `fixed_latency`. If a date, it gives
+#'   the date from which the forecast is actually occurring. If `NULL`, the
+#'   `forecast_date` is determined either via the `fixed_latency`, or is set to
```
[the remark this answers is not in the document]
dsweber2: `fixed_*` are mostly about giving the user easy options, depending on which feature of latency they actually know, rather than having to compute it by hand. My explanation may not be particularly good though, and if you have suggestions for a different interface I'd welcome that.

A lot of it shakes out to simple checking for `NULL` before actually setting a value, since I have to set both `latency` and `forecast_date` anyways. The option just replaces the calculated versions with a fixed one.

After thinking through the logic again, `epi_keys_checked` is always used, either as part of computing the `forecast_date` given a `fixed_latency` or as part of computing the `latency` given a `fixed_forecast_date`, so it doesn't actually get caught up in this.

R/utils-latency.R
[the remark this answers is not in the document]
dsweber2: gotcha, I'll switch them both then

dajmcdon:
@dsweber2 Is this the PR I'm blocking? Is there something in particular I should focus on in review?

dsweber2:
This one and it's parent `adjustAhead` are both more or less waiting on your review. Couldn't remember when you were back, probably should've been a bit more explicit and given you a summary to work with when I thought it was ready.

As far as where to focus, probably the arguments to `layer_add_forecast_date`, `layer_add_target_date`, and the `arx_*` functions, and the tests are where I would focus first if I were reviewing it, but that's pretty generic.

Rough summary of this PR is:
- `layer_add_forecast_date` and `layer_add_target_date` default to using the date specified by `step_adjust_latency`
- adding `adjust_latency` to `arx_args_list`, with a default of `NULL`, to preserve previous behavior. 
- `forecast_date` and `target_date` both inherit values from `step_adjust_latency`'s logic if `adjust_latency` is present. (Just realized I should move this logic to `arx_args_list`, since it's currently in `arx_forecaster` but not `arx_classifier`; will do when no longer traveling)
- added `get_forecast_date_in_layer` to get the right forecast date in the layers rather than the steps
- changes `as_of` to `forecast_date` for `step_adjust_latency`
- swap out some awkward piping that used braces as a quasi-function

In hindsight, there are some changes I made here that I probably should've made in the other PR and rebased onto, sorry about all the `as_of` -> `forecast_date`'s in the diffs.

R/step_adjust_latency.R
```diff
@@ -13,30 +13,41 @@
 #'   to set this manually, as the necessary adjustments will be done for the
 #'   predictors and outcome.
 #' @param method a character. Determines the method by which the
-#'   forecast handles latency. All of these assume the forecast date is the
-#'   `as_of` of the `epi_df`. The options are:
+#'   forecast handles latency. The options are:
```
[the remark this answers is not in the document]
dajmcdon: I think the @details part here should also say something like

> By default, the latency will be determined using the arguments below, but can be set explicitly using either `fixed_latency` or `fixed_forecast_date`

R/step_adjust_latency.R
```diff
@@ -84,8 +107,9 @@ step_adjust_latency <-
              "locf",
              "extend_lags"
            ),
+           epi_keys_checked = c("geo_value"),
            fixed_latency = NULL,
-           fixed_asof = NULL,
+           fixed_forecast_date = NULL,
            default = NA,
            skip = FALSE,
            columns = NULL,
```
dajmcdon, line 115: What does this do? Is it just populated by the tidy selection? You can inherit these from other `step_*` functions. (Same applies to `skip` and `id`)

R/step_adjust_latency.R
```diff
@@ -114,6 +138,10 @@ step_adjust_latency <-
       cli::cli_abort("adjust_latency needs to occur before any `NA` removal,
                       as columns may be moved around")
     }
+    if (!is.null(fixed_latency) && !is.null(fixed_forecast_date)) {
+      cli::cli_abort("Only one of `fixed_latency` and `fixed_forecast_date`
+ can be non-`NULL` at a time!")
+    }
 
     method <- rlang::arg_match(method)
     terms_used <- recipes_eval_select(enquos(...), recipe$template, recipe$term_info)
```
dajmcdon, line 147: `enquos()` is not in the namespace. Need to import or prefix.

R/step_adjust_latency.R
```diff
@@ -267,9 +299,9 @@ print.step_adjust_latency <-
     } else {
       terms <- x$terms
     }
-    if (!is.null(x$as_of)) {
+    if (!is.null(x$forecast_date)) {
```
dajmcdon, line 302: Printing looks very strange. This is the example recipe above:
Screenshot 2024-06-14 at 10 08 19

R/utils-latency.R
```diff
@@ -68,26 +68,26 @@ construct_shift_tibble <- function(terms_used, recipe, rel_step_type, shift_name
...
+#' @param latency `NULL`, int, or vector, as described in `step_epi_latency`
 #' @param sign_shift -1 if ahead, 1 if lag
 #' @return a tibble with columns `column` (relevant shifted names), `shift` (the
 #'   amount that one is shifted), `latency` (original columns difference between
-#'   max_time_value and as_of (on a per-initial column basis)),
+#'   the max `time_value` and `forecast_date` (on a per-initial column basis)),
 #'   `effective_shift` (shift+latency), and `new_name` (adjusted names with the
 #'   effective_shift)
 #' @keywords internal
 #' @importFrom dplyr rowwise %>%
```
dajmcdon, line 80: The pipe is already imported

R/utils-latency.R
```diff
@@ -135,52 +135,95 @@ set_asof <- function(new_data, info) {
...
-      "`as_of` ({(as_of)}) is before the most ",
+      "`forecast_date` ({(forecast_date)}) is before the most ",
       "recent data ({max_time}). Remove before ",
       "predicting."
     ))
   }
   # TODO cover the rest of the possible types for as_of and max_time...
-  if (class(time_values) == "Date") {
-    as_of <- as.Date(as_of)
+  if (class(max_time) == "Date") {
```
dajmcdon, line 177: should be `inherits()`

R/utils-latency.R
```diff
@@ -135,52 +135,95 @@ set_asof <- function(new_data, info) {
   }
   # the source data determines the actual time_values
   # these are the non-na time_values;
-  time_values <- new_data %>%
+  # get the minimum value across the checked epi_keys' maximum time values
+  max_time <- new_data %>%
     select(all_of(original_columns)) %>%
```
dajmcdon, line 140: `select()` isn't in the namespace

R/utils-latency.R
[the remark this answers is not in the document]
dajmcdon: You were going to switch these right?

R/utils-latency.R
```diff
@@ -135,52 +135,95 @@ set_asof <- function(new_data, info) {
...
+get_latency <- function(new_data, forecast_date, column, shift_amount, sign_shift, epi_keys_checked) {
   shift_max_date <- new_data %>%
-    drop_na(all_of(column)) %>%
+    drop_na(all_of(column))
+  # null and "" don't work in `group_by`
+  if (!is.null(epi_keys_checked) && epi_keys_checked != "") {
+    shift_max_date <- shift_max_date %>% group_by(get(epi_keys_checked))
+  }
+  shift_max_date <- shift_max_date %>%
+    summarize(time_value = max(time_value)) %>%
```
dajmcdon, line 195: only `summarise()` is in the namespace, not `summarize()`

R/step_adjust_latency.R
dajmcdon: The default for `id` should be `rand_id("adjust_latency")`

R/utils-latency.R
```diff
@@ -135,52 +135,95 @@ set_asof <- function(new_data, info) {
...
-    max()
-  return(as.integer(sign_shift * (as_of - shift_max_date) + shift_amount))
+    min()
+  return(as.integer(sign_shift * (as.Date(forecast_date) - shift_max_date) + shift_amount))
+}
+
+
+
+#' get the target date while in a layer
+get_forecast_date_in_layer <- function(this_recipe, workflow_max_time_value, new_data) {
```
dajmcdon, line 204: I don't think this function works. `as_of` isn't part of the step anymore?

dajmcdon (changes requested):
I left a number of comments, but I suspect there are some missing tests. Or else I'm doing something wrong locally. Checks fail.

R/arx_forecaster.R
```diff
@@ -119,6 +118,13 @@ arx_fcast_epi_workflow <- function(
   if (!(is.null(trainer) || is_regression(trainer))) {
     cli::cli_abort("{trainer} must be a `{parsnip}` model of mode 'regression'.")
   }
+  # forecast_date is first what they set;
+  # if they don't and they're not adjusting latency, it defaults to the max time_value
+  # if they're adjusting as_of, it defaults to the as_of
+  forecast_date <- args_list$forecast_date %||%
+    if (is.null(args_list$adjust_latency)) max(epi_data$time_value) else attributes(epi_data)$metadata$as_of
```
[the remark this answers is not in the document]
dsweber2: yes, it definitely should be

R/step_adjust_latency.R
dsweber2: I was unsure which of these were necessary boilerplate for all steps, and which were args only the other step would need. Would gladly drop most of the generic steps if possible

dajmcdon: I've been basing it off the instructions here (though they may have changed since I last made one): 
https://www.tidymodels.org/learn/develop/recipes/

R/step_adjust_latency.R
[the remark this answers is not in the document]
dsweber2: > Adapt the model to latent data

For some reason I thought we'd used the word pipeline, in the same vein as scikitlearn, but happy to change to whatever wording.

R/step_adjust_latency.R
dsweber2: oh you mean inherit the documentation probably?

R/utils-latency.R
dsweber2: yep, this is one spot I didn't catch the rename; surprised it didn't cause problems with the layer tests.

R/utils-latency.R
dsweber2: there's a number of other ones, renamed any instance I found for consistency

R/utils-latency.R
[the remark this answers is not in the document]
dsweber2: forgot there were 3 instances instead of just 2

R/step_adjust_latency.R
dsweber2: oh I didn't have any tests for this and changed the format. Surprised printing wasn't throwing errors. I'll let you know when I think this is fixed, leaving open

dsweber2:
> Checks fail.

The tests at least were running locally for me at the last PR. I'm generally not in a habit of running the full checks locally since I expect the remote to handle that (and they really slow down the feedback loop); I guess b/c this is a PR on a PR it isn't running the full checks.

Doing so, looks like its mostly things not being in the namespace. Check now passes locally, with 4 notes (mostly some local files it should ignore, the ubiquitous global `*` definition, and some `:::` quasi-imports).

I've marked as resolved the things I thought were straightforward in being addressed, and left open things I'm still confused on or working through.

R/step_adjust_latency.R
dajmcdon: yes, I was wondering specifically if `columns` is necessary. The documentation makes it look like something set by the user, but it's actually ... not used at all?

vignettes/articles/symptom-surveys.Rmd
dajmcdon, line 1: I suspect these were all OK because there's a `library(dplyr)` at the top.

dajmcdon (approved):

R/step_adjust_latency.R
dsweber2: hmm, yeah `columns` is definitely a cargo-cult argument, I'll drop it. `...` are used to actually restrict the terms used to specific columns (though I don't have a test for that, going to add one to make sure it's working properly).

dajmcdon: That's ok. In many example `step_*()`, columns is an argument. I'm not honestly sure why. Take a look at 

```r
?recipes::step_lag
View(recipes::step_lag)
View(recipes:::step_lag_new)
View(recipes:::prep.step_lag)
```
The columns argument gets populated by `terms` at prep time. I see why they use it in `step_lag_new`, but I don't understand why it is kept as an argument to `step_lag`.

Long story short, they say to use it. But it needs to inherit the documentation. You can use
```r
#' @inheritParams recipes::step_lag
```

PR #339, opened 2024-06-14
Missing rlang imports

dajmcdon:
### Checklist

Please:

see name of branch and commit messages


dsweber2:
weird that `is_true` and `is_logical` only come through as problems in a note whereas `select` was classed as an error. Good to have a fix anyway.

It looks like the only way to make the rest of those notes go away is something like [this](https://forum.posit.co/t/data-pronoun-and-no-visible-binding-for-global-variable/85362), where we replace e.g. `time_value` with `.data$time_value`, which is an unfortunate amount of bloat, if straightforward.

dsweber2:
The styler has some opinions on those vignette changes btw; mostly adding more newlines on top of the ones you added

dsweber2 (approved):

PR #350, opened 2024-06-28
refactor: epi_recipe only accepts epi_df

### LINKED ISSUE #348: `epi_recipe` should only accept `epi_df` and error otherwise
Just had Ryan N run into the uninformative error "should be a data.frame, matrix, etc." when giving an epi_recipe and epi_archive, so he just fed in a tibble and then epi_recipe just silently returned a recipe object, which cryptically rejected any attempt to add an `epi_lag`. We should be explicit about expecting an `epi_df` here and error otherwise. In the future, we can add epi_archive support.

dshemetov:
### Checklist

Please:

      dajmcdon.

Only allows epi_df in epi_recipe and everything else just gives a clear error as such. Open to conversation about this, I don't know if falling back to regular recipes is useful sometimes, I haven't needed that functionality, but if it's not, then this is the behavior I prefer.

- Resolves #348

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: epipredict
 Title: Basic epidemiology forecasting methods
-Version: 0.0.15
+Version: 0.0.16
 Authors@R: c(
     person("Daniel", "McDonald", , "daniel@stat.ubc.ca", role = c("aut", "cre")),
     person("Ryan", "Tibshirani", , "ryantibs@cmu.edu", role = "aut"),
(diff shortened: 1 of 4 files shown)


dajmcdon:
I'm working on a more complicated refactor of this, but the short version is I don't think this is the right behaviour. An epi_recipe should be a subclass of recipe (it isn't currently) and operate on non-epi_df. 

In the meantime, if you want it to warn on `epi_recipe.default()` that's fine, but it should produce a recipe. 

A user who downloads our package shouldn't be prevented from recipe behaviour just because they started with the wrong function.

dshemetov:
@dajmcdon Cool, I switched this change over to a warning instead.

dajmcdon (approved):

PR #355, opened 2024-07-07
bug fixes?

dajmcdon:
### Checklist

Please:

      dajmcdon.

I'm seeing local test failures without these fixes. I'm not sure how they pass on the remote (and they date back some time).

1. `expr()` is not currently imported, so it needs a prefix.
2. The `max_time_value` object in both of these can lose the class. If any of the three options are `NULL`, then `max(NULL, ymd("2021-01-01") = 18628` will be an integer.


dsweber2:
Weird that this is failing on the remote, the tests work fine in my local environment, and it looks like the local check is also passing, specifically running `devtools::check('.')`

dsweber2:
for what it's worth the tests also pass locally on `dev` (running the checks now, we'll see). I'm on ubuntu 22.04

dajmcdon:
I'm guessing possibly different versions of R. I'm on 4.4.1. Can you just directly try the code:

```r
max(NULL, as.Date("2021-01-01")) = 18628
```

dajmcdon:
Note that remote checks fail on this as well.

dsweber2:
Guessing you mean `max(NULL, as.Date("2021-01-01")) == 18628`? That's `TRUE`

I'm also on 4.4..1, confusingly

dsweber2:
isn't the `as.Date` taken care of by the later `coerce_time_type` later on anyways?
https://github.com/cmu-delphi/epipredict/blob/6455828b21d07049fb74e2b7c6238d1309492350/R/layer_add_forecast_date.R#L107

dsweber2:
update for posterity: the problem was recipes moving from 1.0.10 to 1.1.0. The change to address this should be backwards compatible.

dsweber2 (approved):
initial changes made sense.

PR #365, opened 2024-07-29
361 dist quantiles null

### LINKED ISSUE #361: `dist_quantiles()` gives confusing error message given zero `values`
Some potential users flagged this issue that they ran into when trying to use `arx_forecaster()`; there must be something going wrong before the call to `dist_quantiles()` I'm trying to track down, but this issue would be good to fix anyway.

``` r
library(epipredict)
#> Loading required package: epiprocess
#> Registered S3 method overwritten by 'tsibble':
#>   method               from 
#>   as_tibble.grouped_df dplyr
#> 
#> 
#>     filter
#> Registered S3 method overwritten by 'epipredict':
#>   method            from   
#>   print.step_naomit recipes
dist_quantiles(list(), c(0.05, 0.95))
#> Error in `as_list_of()` at epipredict/R/dist_quantiles.R:63:3:
#> ! Can't find common type for elements of `x`.
```

<sup>Created on 2024-07-17 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I assume we'd need to provide the expected `.ptype` to `as_list_of()` or raise a more comprehensible error, and/or contribute a better error message for this case upstream, to `vctrs`.

dajmcdon:
### Checklist

Please:

- [x] Request a review from one of the current epipredict main reviewers

- Resolves #361

### DIFF
--- a/R/dist_quantiles.R
+++ b/R/dist_quantiles.R
@@ -5,6 +5,7 @@ new_quantiles <- function(values = double(), quantile_levels = double()) {
   vec_cast(values, double())
   vec_cast(quantile_levels, double())
   stopifnot(length(values) == length(quantile_levels))
+  if (length(values) == 0L) return(distributional::dist_degenerate(double()))
   stopifnot(!vctrs::vec_duplicate_any(quantile_levels))
   if (is.unsorted(quantile_levels)) {
     o <- vctrs::vec_order(quantile_levels)
@@ -60,6 +61,7 @@ dist_quantiles <- function(values, quantile_levels) {
   values <- as_list_of(values, .ptype = double())
   quantile_levels <- as_list_of(quantile_levels, .ptype = double())
   args <- vec_recycle_common(values = values, quantile_levels = quantile_levels)
+  if (length(args$values) == 0L) return(distributional::dist_degenerate(double()))
   qntls <- as_list_of(map2(args$values, args$quantile_levels, new_quantiles))
   new_vctr(qntls, class = "distribution")
 }
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: epipredict
 Title: Basic epidemiology forecasting methods
-Version: 0.0.17
+Version: 0.0.18
 Authors@R: c(
     person("Daniel", "McDonald", , "daniel@stat.ubc.ca", role = c("aut", "cre")),
     person("Ryan", "Tibshirani", , "ryantibs@cmu.edu", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -52,3 +52,4 @@ Pre-1.0.0 numbering scheme: 0.x will indicate releases, while 0.0.x will indicat
   `...` args intended for `predict.model_fit()`
 - `bake.epi_recipe()` will now re-infer the geo and time type in case baking the
   steps has changed the appropriate values
+- produce length 0 `dist_quantiles()`
--- a/tests/testthat/test-dist_quantiles.R
+++ b/tests/testthat/test-dist_quantiles.R
@@ -24,6 +24,13 @@ test_that("single dist_quantiles works, quantiles are accessible", {
     extrapolate_quantiles(z, c(.3, .7), middle = "linear"),
     new_quantiles(values = c(1, 1.5, 2, 3, 4, 4.5, 5), quantile_levels = 2:8 / 10)
   )
+  # empty values slot results in a length zero distribution
+  # see issue #361
+  expect_length(dist_quantiles(list(), c(.1, .9)), 0L)
+  expect_identical(
+    dist_quantiles(list(), c(.1, .9)),
+    distributional::dist_degenerate(double())
+  )
 })
 
 


R/dist_quantiles.R
```diff
@@ -5,6 +5,9 @@ new_quantiles <- function(values = double(), quantile_levels = double()) {
   vec_cast(values, double())
   vec_cast(quantile_levels, double())
```
brookslogan, line 6: (this is a no-op. `vec_cast` doesn't have side effects)

R/dist_quantiles.R
```diff
@@ -5,6 +5,9 @@ new_quantiles <- function(values = double(), quantile_levels = double()) {
...
   stopifnot(length(values) == length(quantile_levels))
+  if (length(values) == 0L) {
+    return(distributional::dist_degenerate(double()))
+  }
```
brookslogan, line 10: question: is this out of order vs. line 7?

question: why not a 0-length `dist_quantiles`?

R/dist_quantiles.R
dajmcdon: I'm not positive this is the "best" thing to do. It doesn't need to convert. I'm open to suggestions.

R/dist_quantiles.R
dajmcdon: Right. I don't think we should ever hit this line. But I put it in anyway.

R/dist_quantiles.R
```diff
@@ -5,6 +5,9 @@ new_quantiles <- function(values = double(), quantile_levels = double()) {
   vec_cast(values, double())
   vec_cast(quantile_levels, double())
   stopifnot(length(values) == length(quantile_levels))
+  if (length(values) == 0L) {
+    return(distributional::dist_degenerate(double()))
```
brookslogan, line 9: issue: this is the wrong type to come out of `new_quantiles()`; `dist_degenerate(double())` is a `"distribution"`, not an element thereof.  Plus it seems like asking for trouble anyway to have these ill-formed distributions around.

```suggestion
    cli_abort("new_quantiles() requires at least one quantile to be specified.  If you meant to make an empty list of distributions, try dist_quantiles() instead.")
```

R/dist_quantiles.R
```diff
@@ -60,6 +63,9 @@ dist_quantiles <- function(values, quantile_levels) {
   values <- as_list_of(values, .ptype = double())
   quantile_levels <- as_list_of(quantile_levels, .ptype = double())
   args <- vec_recycle_common(values = values, quantile_levels = quantile_levels)
+  if (length(args$values) == 0L) {
+    return(distributional::dist_degenerate(double()))
+  }
   qntls <- as_list_of(map2(args$values, args$quantile_levels, new_quantiles))
```
brookslogan, line 69:
```suggestion
  qntls <- as_list_of(map2(args$values, args$quantile_levels, new_quantiles),
                      .ptype = new_quantiles(NA_real_, 0.5))
```

brookslogan: or
```suggestion
  qntls <- map2(args$values, args$quantile_levels, new_quantiles)
```

R/dist_quantiles.R
```diff
@@ -5,6 +5,9 @@ new_quantiles <- function(values = double(), quantile_levels = double()) {
   vec_cast(values, double())
   vec_cast(quantile_levels, double())
   stopifnot(length(values) == length(quantile_levels))
```
brookslogan, line 7:
```suggestion
  stopifnot(length(values) == length(quantile_levels))
  values <- unname(values)
```
this can prevent `vctrs::vec_cast_common` from producing things like
```
Error:
! Can't convert `..1` <dist_quantiles> to <dist_quantiles>
```
but you'll need to accept/review the snapshot tests.

brookslogan (approved):
- I think `new_quantiles` needs a small fixup to be a bit more consistent/demanding.
- I've suggested an alternative approach that's mostly stylistic, but also involves avoiding some unclear vctrs messages that we might encounter in further development.

R/dist_quantiles.R
brookslogan: I'm not sure.  It certainly seems simpler to think about if we could go ahead and cast.  But I'm not sure if there's a situation where someone really really wants their integer-type/esoteric-yet-still-double-compatible quantiles.

R/dist_quantiles.R
dajmcdon: Going with the first version.

R/dist_quantiles.R
dajmcdon: Will do.

R/dist_quantiles.R
dajmcdon: agreed. Also changing the function defaults to `double(1)` to avoid triggering by default.

dajmcdon:
@brookslogan I think I've addressed all your suggestions, but maybe take a quick look if you can. I also added some code to compute weighted interval score for these distributions (which we need to use in various places). Sorry!

brookslogan (approved):
Left a few more thoughts.  Like the last batch, all sort of preference-based.

R/weighted_interval_score.R
```diff
@@ -0,0 +1,73 @@
...
+#' ) %>%
+#'   mutate(wis = weighted_interval_score(.pred_distn, actual))
+#' preds
+weighted_interval_score <- function(x, actual, ...) {
+  UseMethod("weighted_interval_score")
+}
+
+#' @export
+weighted_interval_score.default <- function(x, actual, ...) {
+  cli_abort("Weighted interval score can only be calculated if `x` has class {.cls distribution}.")
```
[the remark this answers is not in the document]
dajmcdon: A `distribution` is a vector container that can contain 1 or more `dist_quantiles`. But it can contain different types of distributions simultaneously. Like:
``` r
library(distributional)
dstn <- c(dist_normal(), dist_quantiles(1:4, 1:4/5))
```

PR #379, opened 2024-09-12
Retain epi df class

### LINKED ISSUE #376: Steps sometimes destroy the `epi_df`
* `step_epi_naomit()` is one such. `step_naomit()` calls `tibble::as_tibble()` at the end.

Consider reconstructing the `epi_df` after each step inside the `bake/prep` loop.

dajmcdon:
### Checklist

Please:

- Resolves #376 to the extent possible


dsweber2:
How does this effect https://github.com/cmu-delphi/epipredict/pull/370, deprecating the epi_recipe?

dajmcdon:
This is a short term fix. That one is far more major (but still won't quite fix this problem). I'll add the no decay stuff there as well.

dshemetov (approved):
Seems like a nice-to-have partial fix.

Note: this flag activates a hook in epiprocess::as_tibble.epi_df which attempts to preserve the metadata in the returned tibble. Naturally, appending metadata to an existing class comes with no guarantees that it will be preserved or that it won't interfere with other aspects of the class, but (a) it seems unlikely that we will have issues here, (b) very very hard to understand the downstream consequences without actually trying this, so I think we should go ahead with it.

Need to fix: version number in DESCRIPTION.

PR #381, opened 2024-09-16
Adjust ahead djm

dajmcdon:


dajmcdon:
@dsweber2 This is the idea I had in mind. It probably breaks checks. But perhaps we can iterate on this.

R/step_adjust_latency.R
```diff
@@ -310,21 +303,26 @@ prep.step_adjust_latency <- function(x, training, info = NULL, ...) {
 bake.step_adjust_latency <- function(object, new_data, ...) {
   if (!inherits(new_data, "epi_df") || is.null(attributes(new_data)$metadata$as_of)) {
     new_data <- as_epi_df(new_data)
-    attributes(new_data)$metadata <- object$metadata
+    attributes(new_data)$metadata <- object$metadata # DJM: why? detected above?
```
dsweber2, line 306: this is a patch to reinstate the metadata if its removed

R/utils-latency.R
```diff
@@ -28,6 +28,34 @@ construct_shift_tibble <- function(terms_used, recipe, rel_step_type, shift_name
...
+    method <- latency_step$method
+    if (method == "extend_ahead") {
+      loc <- which(step_types == "epi_ahead")
+      sign_shift <- -1
+    } else if (method == "extend_lags") {
+      loc <- which(step_types == "epi_lag")
+      sign_shift <- 1
+    }
+    if (method != "locf") {
+      for (s in seq_along(loc)) {
```
dsweber2, line 47:
```suggestion
      for (s in loc) {
```

R/epi_recipe.R
```diff
@@ -569,6 +569,10 @@ bake.epi_recipe <- function(object, new_data, ..., composition = "epi_df") {
     }
     composition <- "tibble"
   }
+
+  ## Latency checks/adjustments
+  object <- adjust_recipe_latency_before_bake(object)
```
dsweber2, line 574: So because this is happening during bake, the `term_info` has the wrong column names; the specific error is
```r
Browse[1]> Error in `vctrs::vec_locate_matches()`:
! Each value of `needles` must have a match in `haystack`.
✖ Location 5 of `needles` does not have a match.
```
. Not sure if it'd be easier to adjust here, or move this inside the prep step of `step_epi_shift`

R/utils-latency.R
```diff
@@ -286,33 +318,38 @@ check_interminable_latency <- function(dataset, latency_table, target_columns, f
...
+get_latency_table <- function(training, columns, forecast_date, latency,
+                              sign_shift, epi_keys_checked, info, terms) {
   if (is.null(columns)) {
     columns <- recipes_eval_select(terms, training, info)
   }
   # construct the latency table
-  latency_table <- names(training)[!names(training) %in% key_colnames(training)] %>%
-    tibble(col_name = .)
+  latency_table <- tibble(col_name = names(training)) %>%
+    filter(col_name %nin% key_colnames(training))
```
dsweber2, line 328: honestly surprised this isn't already a thing. much cleaner

R/epi_recipe.R
dajmcdon: Should be fixed by 28adec2

dsweber2:
Alright done! Checks pass locally, even if they won't run here. Side bonus, this adds a label for trained steps that have had their ahead/lag modified and includes the modified version instead.

dajmcdon:
I'm ok if you want to merge this now and continue to iterate on the big PR?

Two things:
1. The edits to `prep.epi_recipe()` will not persist for very long. The intention is for this to look like https://github.com/cmu-delphi/epipredict/blob/6530c7e3ca75c7f6a2f76f8c61f9a5c5d537452e/R/epi_recipe.R#L226.
2. There's some attribute setting/tracking. Is that necessary? My intention here was to avoid doing that.

dsweber2:
In that case the changes you're asking for in this PR are impossible. We're adding a feature (metadata) that's unsupported by base recipes, so we need to modify either the `prep` or `bake` functions for our use case.

dsweber2:
re 2, we could add logic that detects if the step is a shift and attaches the metadata to it, but I think that ends up being more fragile/error prone than adding it to the epi_df. It moves the logic for a particular step into `prep.epi_recipe`, rather than into that step.

dajmcdon:
I think we want the logic in `prep.epi_recipe()`. That way, it can be processed either before or after dispatch to `prep.recipe()` (this is simlar to the `adjust_recipe_latency_before_bake()` logic). This avoids having to duplicate the guts `prep.recipe` with only minor changes. 

It also means that we can avoid setting attributes. Because it's in `prep.epi_recipe()` it can access the entire recipe and directly modify the `ahead/lag` steps if needed.

dsweber2:
I see, in that case it's still technically possible, if quite a headache; probably moreso than keeping the custom loop would be.

If we do it after, we have to throw out and recompute `var_info`, `term_info` , and `last_term_info`. Any steps that depend on `training` having the right shifts after will probably be wrong (`step_naomit` comes to mind), so we should recompute all of those. This means restarting the prep/bake process at `step_adjust_latency` so we're doing the second half of the loop anyway.

To do it beforehand, you have to run through all the previous steps to get the state to be correct to know which columns exist and what their latencies are, so we'd be doing the first half of the for loop, and then applying the logic.

Both of these sound like significantly more work than maintaining a lightly customized version of the for-loop of `prep.epi_recipe`, since both effectively require doing that anyway to write a partial execution of that loop, and then dealing with the resulting logic. Looking at [the blame for prep.recipe](https://github.com/tidymodels/recipes/blame/d84bf92885ab98da25f643339c570070743f0027/R/recipe.R#L391), the function seems fairly stable at this point. It would certainly be nice if they had made their function more modular so it was calling functions instead of duplicating exact logic.

dajmcdon:
OK, in that case, I'd rather avoid modifying the prep/bake internals and return to the old implementation based on attribute munging. Do you want to close this and grab off anything useful for the other branch?

PR #416, opened 2024-10-21
doc: update sliding vignette

dshemetov:
### Checklist

Please:

      dajmcdon.
- [ ] Make sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
- [ ] Describe changes made in NEWS.md, making sure breaking changes

Updates the sliding vignette, brings it out of articles to the dropdown, and renames it to backtesting. Lines the vignette up to follow along with the rewritten [vignette in epiprocess](https://cmu-delphi.github.io/epiprocess/articles/epi_archive.html), but instead use the `arx_forecaster`.

- Resolves #{issue number}


DESCRIPTION
```diff
@@ -36,6 +36,7 @@ Imports:
     ggplot2,
     glue,
     hardhat (>= 1.3.0),
+    here,
```
dajmcdon, line 39:
```suggestion
```

This isn't necessary except for ease in local vignette rendering (possibly). See below.

DEVELOPMENT.md
```diff
@@ -13,35 +11,69 @@ devtools::build() # build package
...
 ## Developing the documentation site
 
-The [documentation site](https://cmu-delphi.github.io/epipredict/) is built off of the `main` branch. The `dev` version of the site is available at https://cmu-delphi.github.io/epipredict/dev.
-
-The documentation site can be previewed locally by running in R
-
-```r
-pkgdown::build_site(preview=TRUE)
-```
+Our CI builds two version of the documentation:
```
dajmcdon, line 24:
```suggestion
Our CI builds two versions of the documentation:
```

pkgdown-watch.r
dajmcdon, line 1: I don't think that this is necessary in this package (rather than elsewhere). This is developer use only, correct? 

I could imagine it in `/inst`, perhaps. RStudio will do this for you, so it seems more likely to be specific to your workflow.

vignettes/.gitignore
```diff
@@ -1,3 +1,4 @@
 *.html
 *_cache/
 *.R
+!_common.R
```
dajmcdon, line 4:
```suggestion
!_common.R

```

vignettes/backtesting.Rmd
```diff
@@ -0,0 +1,460 @@
...
+title: "Accurately backtesting forecasters"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Accurately backtesting forecasters}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r setup, include = FALSE}
+source(here::here("vignettes", "_common.R"))
```
dajmcdon, line 11: Since this is used only for vignette building, the `here` package would go under `Suggests` in DESCRIPTION. 

However, you don't need the path like this. When built, vignettes access this path. See: https://r-pkgs.org/vignettes.html#filepaths

vignettes/backtesting.Rmd
```diff
@@ -0,0 +1,460 @@
...
+```
+
+```{r pkgs}
```
dajmcdon, line 14:
```suggestion
```{r pkgs, include=FALSE, echo=TRUE}
```

vignettes/backtesting.Rmd
dajmcdon: To iterate on building this vignette locally, you can do:

`devtools::build_rmd("vignettes/backtesting.Rmd")` which rebuilds the package, then renders the vignette.

vignettes/backtesting.Rmd
```diff
@@ -0,0 +1,460 @@
...
+last date, and then run the forecaster.
+
+```{r}
+# The .versions argument selects only the last version in the archive and
+# produces a forecast only on that date.
+forecasts <- archive_cases_dv_subset %>%
+  epix_slide(
+    ~ arx_forecaster(
+      .x,
+      outcome = "percent_cli",
```
dajmcdon, line 141: Do you really want `percent_cli` as the outcome? I think you want to predict cases, right?

vignettes/backtesting.Rmd
```diff
@@ -0,0 +1,460 @@
...
+# Join the forecasts with with the latest data at the time of the forecast to
+# compare. Since `percent_cli` data has a few days of lag, we use `tidyr::fill` to
+# fill the missing values with the last observed value.
+forecasts %>%
+  inner_join(
+    archive_cases_dv_subset %>%
+      epix_as_of(archive_cases_dv_subset$versions_end) %>%
+      group_by(geo_value) %>%
+      tidyr::fill(percent_cli),
+    by = c("geo_value", "forecast_date" = "time_value")
```
dajmcdon, line 155: I don't think you want to join `forecast_date` to `time_value`. The target date is `forecast_date + 7L` right?

vignettes/backtesting.Rmd
```diff
@@ -0,0 +1,460 @@
...
+  ) %>%
+  select(geo_value, forecast_date, .pred, .pred_distn, percent_cli)
+```
+
+The resulting epi_df now contains two new columns: `.pred` and `.pred_distn`,
+corresponding to the point forecast (median) and the quantile distribution
+containing our requested quantile forecasts (in this case, 0.05 and 0.95)
+respectively. The forecasts look reasonable and in line with the data. The point
+forecast is close to the last observed value (joined in the table above as
+`percent_cli`), and the uncertainty bands are wide enough to capture the
```
dajmcdon, line 165: How are you evaluating the quality of the bands?

dajmcdon (changes requested):

vignettes/backtesting.Rmd
dshemetov: Vignettes having access to their own folder is a really unfortunate convention 😠 I prefer to iterate on vignette code in my REPL, not by rebuilding the whole vignette every time (and I know that caching speeds things up, but a built vignette doesn't let me interact with the objects like a REPL). I'm surprised this isn't an issue for RStudio users.

Not gonna waste time on figuring out a workaround, so I'll just do the relative path thing.

vignettes/backtesting.Rmd
dshemetov: Yea, it's strange to forecast `percent_cli`, but it has the more drastic revision history. Cases from JHU barely have any revisions. This example could definitely be improved.

vignettes/backtesting.Rmd
dshemetov: Just whether the values are close, it's just meant to be a simple sanity check.

vignettes/backtesting.Rmd
dajmcdon: This is a modification of the former `sliding.Rmd` right? I think it's important to have a believable forecast task. Predicting cli with something that almost certainly happens _after_ is not a great illustration.

vignettes/backtesting.Rmd
dajmcdon: The uncertainty bands aren't visible.

vignettes/backtesting.Rmd
dshemetov: It's because I was forecasting on the last date of the dataset, so I didn't have forecast_date + 7L to compare. I'm going to change this example to forecasting from last date - 14, so we have true data to compare.

vignettes/backtesting.Rmd
dshemetov: Fixing that right now

vignettes/backtesting.Rmd
dshemetov: The previous `sliding.Rmd` seemed a bit unfocused to me: it seemed to both be trying to demonstrate version vs unversioned backtesting, but when plotting only compared the performance different engines for some reason. Like I said, forecasting cases make for a bad version vs unversioned example because there's no visible revisions, unlike percent_cli. 

The forecasts are not unbelievably good, you can see similar results in the new [epiprocess vignette](https://cmu-delphi.github.io/epiprocess/dev/articles/epi_archive.html#backtesting-a-simple-autoregressive-forecaster) using `prob_ar()`. How do you reason that we're getting future information for forecasting? We're using epix_slide, so all our data is versioned.

dajmcdon: percent_cli is constructed to be a leading indicator. We get that number before the person is tested and the test is reported to PH. So CLI precedes cases in time. Like predicting cases with deaths. The time arrow is backward. If you want to use cli as the outcome, we shouldn't use cases as a predictor.

dshemetov: I see, makes sense. How about I just drop the exogenous variable for now and we can find one later?

dajmcdon (approved):

inst/pkgdown-watch.r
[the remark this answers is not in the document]
dshemetov: It took me hours to find the iterative documentation workflow provided by this script. Figured I'd save that for other devs 🤷

PR #433, opened 2025-01-31
380 add forecast date recipe

### LINKED ISSUE #380: add `forecast_date` as a first-class part of the recipe object
Currently the forecast date is scattered across several functions. After https://github.com/cmu-delphi/epipredict/pull/370, we should include it as a part of recipe construction in the first place, so we don't have multiple places where it might be added that might contradict eachother.

dajmcdon:
### Checklist

Please:

      dajmcdon.

Added `reference_date` to the epi recipe. I think this is likely a better terminology. The hope is to use these for nowcasts and backcasts, so the thinking here is to make downstream computations relative to the `reference` in all cases. It's not specific only to forecasts.

_Note:_ Downstream steps (such as `step_adjust_latency()`) can be updated in separate PRs.

- Resolves #380


dsweber2 (approved):
Possible problem down the road: Roni is talking about switching `time_value` for `reference_date` in epidata, so we may want to too, which would lead to some confusing name collision. Probably worth going ahead with this anyways.

Just confirming, no steps/layers are currently using this right? E.g. `layer_add_forecast_date`?

dajmcdon:
1. Correct. This is not currently being used by any layers or steps downstream. We can make those changes later.
2. On the `reference_date` name, I considered `forecast_date` (as in the issue name), but that seems poor relative to other non-forecasting tasks. I also thought about `training_date`, but the model technically isn't trained until it is `fit()`, so that doesn't seem right either. Maybe `model_creation_date`? I like `reference_date` because it's an important piece of information to which other methods may refer. But I'm open to alternatives.

dsweber2:
`reference_date` would definitely be the best option without the collision. `model_date` might work; I think `model_creation_date` may actually be pointing to the day when the model is fit rather than the forecast date (mostly comes up in backtesting). Otherwise `forecast_date` seems clear enough even if it might not be an exact fit for backfill projection and nowcasting. Possibly `prediction_date`?

Issue #434, opened 2025-02-07
`extrapolate_quantiles` and thresholding


dsweber2:
Discovered in #431, if you try to extrapolate quantiles for predictions which have been thresholded, you can get quantiles outside of those thresholds. For example, if you have a quantile level of `.1` which has value `0`, then `.05` will be negative, even if the quantile is negative.

Ways I can think to deal with this:
1. include the support in the distribution, probably as an interval, e.g. `c(0,Inf)`, and use that in `quantile_extrapolate`. Unsure if the types will play nicely with this one.
2. Add limit options to get passed to `quantile_extrapolate`. Seems like it will require the user to know that they have thresholds, whereas the first option will build it into the result as part of `layer_threshold`.
3. Just let it be and let the user figure it out by mentally thresholding. Seems not ideal.

dajmcdon:
@dsweber2 Is this still relevant or effectively closed by #435?

dsweber2:
The bug is still live, it's just not on the default path. #435 just decreased the urgency

dajmcdon:
Can you give an example of where it happens? I'm OK with option 2.

dajmcdon:
[It's pretty undesirable to threshold and then extrapolate. So, we shouldn't be doing this anywhere. And ideally, the user wouldn't either. No matter what the "fix" it will be demonstrably worse than doing it in the other order.]

dsweber2:
If they change the default `quantile_levels` in the original example, they can reproduce the original behavior
```r
forecast_date <- as.Date("2021-08-01")
two_week_ahead <- arx_forecaster(
  covid_case_death_rates |>
    filter(time_value <= forecast_date),
  outcome = "death_rate",
  trainer = quantile_reg(),
  predictors = c("death_rate"),
  args_list = arx_args_list(
    lags = list(c(0, 7, 14)),
    ahead = 14,
    quantile_levels = c(0.2, 0.3, 0.5, 0.7, 0.8)
  )
)
```

dajmcdon:
Now I see. In that case, I'm leaning toward either:

1. Document the extrapolation behaviour and note that user deciding to do something that they shouldn't may lead to results they don't want.
2. Remove the ability to extrapolate completely. Allow it to do imputation in the middle, but refuse to handle the tails.

dsweber2:
how does this interact with https://github.com/cmu-delphi/epipredict/pull/442 by the way?

dajmcdon:
[...]

PR #431, opened 2025-01-23
Docs overhaul

dsweber2:
### Checklist

Please:

- [X] Make sure this PR is against "dev", not "main".
- [x] Request a review from one of the current epipredict main reviewers:
      dajmcdon.
- [ ] Make sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
      Always increment the patch version number (the third number), unless you are
      making a release PR from dev to main, in which case increment the minor
      version number (the second number).
- [ ] Describe changes made in NEWS.md, making sure breaking changes
      (backwards-incompatible changes to the documented interface) are noted.
      Collect the changes under the next release number (e.g. if you are on
      0.7.2, then write your changes under the 0.8 heading).
- [X] Consider pinning the `epiprocess` version in the `DESCRIPTION` file if
  - You anticipate breaking changes in `epiprocess` soon
  - You want to co-develop features in `epipredict` and `epiprocess`

### Change explanations for reviewer

Draft ready for review:
- [X] Landing Page
- [x] Getting Started
- [x] Customized Forecasters
- [x] Reference
  - [x] Using the add/update/remove and adjust functions
  - [x] Smooth Quantile regression
- [x] preprocessing and models examples
- [x] backtesting forecasters

### Magic GitHub syntax to mark associated Issue(s) as resolved when this is merged into the default branch

- [ ] Resolves https://github.com/cmu-delphi/forecasting-team-meta/issues/26
- [ ] Resolves https://github.com/cmu-delphi/epipredict/issues/264
- [ ] Resolves https://github.com/cmu-delphi/epipredict/issues/261
- [ ] Resolves https://github.com/cmu-delphi/epipredict/issues/256
- [x] Resolves https://github.com/cmu-delphi/epipredict/issues/320
- [x] Resolves #232

PR #435, opened 2025-02-07
Canned median quantile

dsweber2:
### Checklist

Please:

- [ ] Request a review from one of the current epipredict main reviewers:
      dajmcdon.

This is a minor patch to add the median as a fit quantile as a default, which deals with some problems caught in #431. The median can change radically after thresholding if there are too few quantiles, so this fixes the median by making sure the median is also fit and included in the distribution. It doesn't solve extrapolation, which is left open as an issue.

Edit:
this is a patch more or less, with a more principled fix discussed in #434

### DIFF
--- a/R/autoplot.R
+++ b/R/autoplot.R
@@ -236,7 +236,7 @@ plot_bands <- function(
     alpha = 0.6,
     linewidth = 0.05) {
   innames <- names(predictions)
-  na_levels <- length(levels)
+  n_levels <- length(levels)
   alpha <- alpha / (n_levels - 1)
   # generate the corresponding level that is 1 - level
   levels <- (1 - levels) / 2
--- a/R/autoplot.R
+++ b/R/autoplot.R
@@ -39,9 +39,7 @@ ggplot2::autoplot
 #'   step_epi_naomit()
 #'
 #' f <- frosting() %>%
-#'   layer_residual_quantiles(
-#'     quantile_levels = c(.025, .1, .25, .75, .9, .975)
-#'   ) %>%
+#'   layer_residual_quantiles() %>%
 #'   layer_threshold(starts_with(".pred")) %>%
 #'   layer_add_target_date()
 #'
--- a/R/autoplot.R
+++ b/R/autoplot.R
@@ -85,7 +85,7 @@ NULL
 #' @rdname autoplot-epipred
 autoplot.epi_workflow <- function(
     object, predictions = NULL,
-    .levels = c(.5, .8, .95), ...,
+    .levels = c(.5, .8, .9), ...,
     .color_by = c("all_keys", "geo_value", "other_keys", ".response", "all", "none"),
     .facet_by = c(".response", "other_keys", "all_keys", "geo_value", "all", "none"),
     .base_color = "dodgerblue4",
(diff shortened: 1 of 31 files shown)


dajmcdon:
It's not entirely obvious to me that:
1. The bug shown is fixed by this change.
2. This change doesn't introduce other unexpected behaviour.

Can you explain a bit why this should result in the intended behaviour? It seems to me that there are at least two things going on:
1. Autplot has a default with a 95% interval, but flatline doesn't have the same default (neither do other canned forecasters). This means that it tries to extrapolate them (by default), but the extrapolation can go negative (possibly suggesting that a fix should be made to the extrapolation implementation).
2. The flat point forecast clearly is much different from the centre of the plotted intervals. I think the intervals there are the problem.

dajmcdon:
FYI: there's a missing `dplyr::` on line 186 of `autoplot.R`

R/autoplot.R
```diff
@@ -236,19 +236,20 @@ plot_bands <- function(
     alpha = 0.6,
     linewidth = 0.05) {
   innames <- names(predictions)
-  n <- length(levels)
-  alpha <- alpha / (n - 1)
-  l <- (1 - levels) / 2
-  l <- c(rev(l), 1 - l)
+  na_levels <- length(levels)
```
dshemetov, line 239: typo?
```suggestion
  n_levels <- length(levels)
```

dsweber2:
You're right that it's the quantiles that are wrong. The problem is the interpolated quantiles get wrecked when one of the two quantiles gets adjusted by thresholding, making the median of the quantiles no longer equal to the point prediction.

This is a fix for the median of the quantiles not being equal to the point prediction. It doesn't fix the fact that there are negative extrapolated quantiles, I'll add your suggestion for that in a minute.

compare 
## the case with only 2 extreme quantiles
```
forecast_date <- as.Date("2021-08-01")
used_locations <- c("ca", "ma", "ny", "tx")
all_flatlines <- lapply(
  seq(0, 28, by = 7),
  \(days_ahead) {
    flatline_forecaster(
      covid_case_death_rates |>
        filter(time_value <= forecast_date, geo_value %in% used_locations),
      outcome = "death_rate",
      args_list = flatline_args_list(
        ahead = days_ahead,
        quantile_levels = c(0.05, 0.95)
      )
    )
  }
)
# same plotting code as in the arx multi-ahead case
workflow <- all_flatlines[[1]]$epi_workflow
results <- purrr::map_df(all_flatlines, ~ `$`(., "predictions"))
results %>% filter(target_date == max(target_date))
autoplot(
  object = workflow,
  predictions = results,
  plot_data = covid_case_death_rates |> filter(geo_value %in% used_locations, time_value > "2021-07-01")
)
```
ex2
## the case with a median quantile
```
forecast_date <- as.Date("2021-08-01")
used_locations <- c("ca", "ma", "ny", "tx")
all_flatlines <- lapply(
  seq(0, 28, by = 7),
  \(days_ahead) {
    flatline_forecaster(
      covid_case_death_rates |>
        filter(time_value <= forecast_date, geo_value %in% used_locations),
      outcome = "death_rate",
      args_list = flatline_args_list(
        ahead = days_ahead,
        quantile_levels = c(0.05, 0.95)
      )
    )
  }
)
# same plotting code as in the arx multi-ahead case
workflow <- all_flatlines[[1]]$epi_workflow
results <- purrr::map_df(all_flatlines, ~ `$`(., "predictions"))
results %>% filter(target_date == max(target_date))
autoplot(
  object = workflow,
  predictions = results,
  plot_data = covid_case_death_rates |> filter(geo_value %in% used_locations, time_value > "2021-07-01")
)
```
with the same but `quantile_levels = c(0.05, 0.5, 0.95)`
ex1

dajmcdon (approved):

PR #437, opened 2025-02-11
Hotfix growth rate

dajmcdon:
### Checklist

Please:

      dajmcdon.

- This changes is necessary for CI which builds from epiprocess >= 0.10.4. (following cmu-delphi/epiprocess#595)
- Note also the minor related issue cmu-delphi/epiprocess#605


R/step_growth_rate.R
```diff
@@ -46,11 +49,11 @@ step_growth_rate <-
            horizon = 7,
            method = c("rel_change", "linear_reg"),
            log_scale = FALSE,
+           na_rm = TRUE,
            replace_Inf = NA,
            prefix = "gr_",
            skip = FALSE,
-           id = rand_id("growth_rate"),
-           additional_gr_args_list = list()) {
```
dsweber2, line 53: I'm wondering why you're dropping `additional_gr_args_list`; `growth_rate` appears to still have a `params` option. Were you planning on not passing those through?

dsweber2 (commented):
I was just running into this over on #431; I'm surprised the tests for `step_growth_rate` were passing before explicitly specifying which argument is `x` and which is `y` for `growth_rate`.

R/step_growth_rate.R
dajmcdon: Currently, the only supported methods are `rel_change` and `linear_reg`. The others can't be used for forecasting (without major changes to lots of things). Those methods don't need/allow any additional arguments through that list.

dajmcdon: I had local tests passing until I opened an unrelated PR. Locally, I had epiprocess 0.10.0. But if you're on current epiprocess dev (>=0.10.4) then the tests all fail. (and they also fail on CI after it installs from the default branch)

dsweber2: right, makes sense. Thought I had local tests passing on epiprocess 0.10.4, but I must've just been confused. Thanks for the clarification.

NEWS.md
```diff
@@ -11,6 +11,8 @@ Pre-1.0.0 numbering scheme: 0.x will indicate releases, while 0.0.x will indicat
   `data(<dataset name>)`, but can be accessed with 
   `data(<dataset name>, package = "epidatasets")`, `epidatasets::<dataset name>` 
   or, after loading the package, the name of the dataset alone (#382).
+- Addresses upstream breaking changes from cmu-delphi/epiprocess#595 (`growth_rate()`). 
+  Corresponding arguments in `step_growth_rate()` have been removed.
```
brookslogan, line 15:
```suggestion
  `step_growth_rate()` has lost its `additional_gr_args_list` argument and now
  has an `na_rm` argument.
```

brookslogan (approved):
Looks like it should work.  I'm getting test failures locally from test-bake-method.R and test-population_scaling.R but they appear to be unrelated.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: epipredict
 Title: Basic epidemiology forecasting methods
-Version: 0.1.8
+Version: 0.1.9
```
dshemetov, line 3:
```suggestion
Version: 0.1.10
```

PR #442, opened 2025-02-25
New hardhat quantiles

dajmcdon:
### Checklist

Please:

- [x] Make sure this PR is against "dev", not "main".
- [x] Request a review from one of the current epipredict main reviewers:
      dajmcdon.
- [x] Make sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
      Always increment the patch version number (the third number), unless you are
      making a release PR from dev to main, in which case increment the minor
      version number (the second number).
- [x] Describe changes made in NEWS.md, making sure breaking changes
      (backwards-incompatible changes to the documented interface) are noted.
      Collect the changes under the next release number (e.g. if you are on
      0.7.2, then write your changes under the 0.8 heading).

### Change explanations for reviewer

This PR replaces `dist_quantiles()`, a custom distribution related to the `distributional` package, with `quantile_pred()` from the `hardhat` package. This is one step in off-loading some custom components to core `tidymodels` functionality.

PR #443, opened 2025-02-26
hardhat quantiles

### LINKED ISSUE #434: `extrapolate_quantiles` and thresholding

dajmcdon:
### Checklist

Please:

      dajmcdon.

This PR replaces dist_quantiles(), a custom distribution related to the distributional package, with quantile_pred() from the hardhat package. This is one step in off-loading some custom components to core tidymodels functionality.

Replaces #442


dsweber2 (approved):
Generally lgtm, very minor notes and added a test for pivot_longer with more than 1 row.

PR #444, opened 2025-02-26
440 epiprocess imports

### LINKED ISSUE #440: Move epiprocess from Depends: to Imports:
If it's under Depends: then we attach epiprocess even if user doesn't want, bloating what's available in global namespace plus causing some conflicts, e.g., with filter, though that probably shouldn't be happening anyway https://github.com/cmu-delphi/epiprocess/issues/606.

dajmcdon:
### Checklist

Please:

- Resolves #440 
- Note that the version number conflicts with #443 
- A few reexports added from tibble, dplyr, tidyr (and epiprocess)
- Most of the file changes are just adjustments wrt reexports in examples.


vignettes/backtesting.Rmd
```diff
@@ -166,7 +167,7 @@ forecasts %>%
     doctor_visits %>%
       epix_as_of(doctor_visits$versions_end) %>%
       group_by(geo_value) %>%
-      tidyr::fill(percent_cli),
+      fill(percent_cli),
```
brookslogan, line 170: This doesn't seem to do what was intended.  Maybe some sort of `complete` was intended beforehand?

```r
waldo::compare(
doctor_visits %>%
      epix_as_of(doctor_visits$versions_end) %>%
      group_by(geo_value) %>%
      fill(percent_cli)
doctor_visits %>%
      epix_as_of(doctor_visits$versions_end) %>%
      `attr<-`("sorted", NULL) %>%
      `attr<-`(".internal.selfref", NULL) %>%
      group_by(geo_value)
)
#> ✔ No differences
```
It does unset some attrs that shouldn't be there... that's [epiprocess#561](https://github.com/cmu-delphi/epiprocess/issues/561)

NAMESPACE
```diff
@@ -160,6 +162,7 @@ export(extract_argument)
 export(extract_frosting)
 export(extract_layers)
 export(extrapolate_quantiles)
+export(filter)
```
brookslogan, line 165: musing: this does introduce a conflict with stats::filter, but I guess eventually we should be providing our own convolution utilities.

brookslogan (approved):
Changes look good; have a couple of non-blocking musings.

vignettes/backtesting.Rmd
dajmcdon: I'm not sure where this was introduced. Hopefully it gets handled in #431 @dsweber2

NAMESPACE
dajmcdon: Yeah, that will happen anytime dplyr is used. This would have happened before anyway since epiprocess was in Depends and it exports `filter()`.

brookslogan: Hmm... I'm not sure why we export filter from epiprocess; shouldn't be necessary just to provide an S3 method.

brookslogan: We could make this happen only when user attaches dplyr themselves.  I guess if users are using it a lot though then it might make sense to re-export.  It's just that convolution does seem somewhat popular with potential users.

PR #445, opened 2025-02-27
Leap time handling

dsweber2:
### Checklist

Please:

- [X] Request a review from one of the current epipredict main reviewers:
      dajmcdon.

So weeks and days both have to deal with inconsistent numbers of days. What this does:
1. separate out the "weird" day/week into a  separate bin, labeled `999` and only use up to the "normal" number of days/weeks (so 365/52)
2. if the window goes over the boundary (so includes both week 52 and week 1, or both day 59 and day 60), then add in the data from 
3. if the window is around the leap day/week, include the `window` days before and the window days after (so days 57, 58, 59, leap_day, 60, 61, 62)

- Resolves #{issue number}


R/step_climate.R
```diff
@@ -341,12 +350,63 @@ roll_modular_multivec <- function(col, .idx, weights, aggr, window_size, modulus
...
+    !lubridate::leap_year(time_value),
+    lubridate::yday(time_value),
+    ifelse(
+      (lubridate::month(time_value) == 2) & lubridate::day(time_value) == 29,
+      999,
+      lubridate::yday(time_value) - (lubridate::month(time_value) > 2)
+    )
+  )
+}
+#' epiweek, but it assigns week 53 the value of -1 instead so it mirrors the assignments in yday_leap
```
dajmcdon, line 403: Do you mean this to say `-1` rather than `999`?

dajmcdon (changes requested):
Some suggestions, but I think there's a logic error when rolling with the `iter`. (Probably was there before as well)

R/step_climate.R
```diff
@@ -341,12 +350,63 @@ roll_modular_multivec <- function(col, .idx, weights, aggr, window_size, modulus
...
+    # +1 from 1-indexing
+    entries <- ((iter - window_size):(iter + window_size) %% modulus)
     entries[entries == 0] <- modulus
+    # note that because we are 1-indexing, we're looking for indices that are 1
+    # larger than the actual day/week in the year
+    if (modulus == 365) {
+      # we need to grab just the window around the leap day on the leap day
+      if (iter == 366) {
+        entries <- ((59 - window_size):(59 + window_size - 1) %% modulus)
+        entries <- c(entries, 366)
```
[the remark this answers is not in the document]
dsweber2: Yes I suppose so, though it requires `window_size = 60` to become relevant

R/step_climate.R
```diff
@@ -341,12 +350,63 @@ roll_modular_multivec <- function(col, .idx, weights, aggr, window_size, modulus
     tidyr::nest(data = c(col, weights), .by = .idx)
   out <- double(nrow(tib))
   for (iter in seq_along(out)) {
-    entries <- (iter - window_size):(iter + window_size) %% modulus
+    # +1 from 1-indexing
+    entries <- ((iter - window_size):(iter + window_size) %% modulus)
```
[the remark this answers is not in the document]
dsweber2: Less than a full year of data is definitely a super pathological case. Switching it to matching `.idx` values rather than assuming the rows are right does seem to work though. I'm including a test in the next commit

R/step_climate.R
dsweber2: yeah I initially wrote this using `-1` as the flag, but having the value at the end rather than the beginning proved somewhat useful.

R/step_climate.R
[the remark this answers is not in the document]
dajmcdon: Is this resolved? (I don't think this is pathological at all. Giving flu seasons, August to April, every year for multiple years and cutting out the summer would break it.)

R/step_climate.R
```diff
@@ -335,18 +345,75 @@ print.step_climate <- function(x, width = max(20, options()$width - 30), ...) {
...
+      purrr::list_rbind(tib |> filter(.idx %in% entries) |> pull(data)),
       aggr(col, weights)
     )
   }
-  tibble(.idx = unique(tib$.idx), climate_pred = out)
+  tibble(.idx = unique(tib$.idx), climate_pred = out[seq_len(nrow(tib))])
+}
+
+
+#' a function that assigns Feb 29th to -1, and aligns all other dates the same
```
dajmcdon, line 394:
```suggestion
#' a function that assigns Feb 29th to 999, and aligns all other dates the same
```

R/step_climate.R
```diff
@@ -335,18 +345,75 @@ print.step_climate <- function(x, width = max(20, options()$width - 30), ...) {
...
+        entries <- c(entries, 999)
+      } else if ((52 %in% entries) || (1 %in% entries)) {
+        # if we're on the year boundary for weekly data, we need to add in the
+        # leap week data (which is the extra week at the end)
+        entries <- c(entries, 999)
+      }
+    }
     out[iter] <- with(
-      purrr::list_rbind(tib$data[entries]),
+      purrr::list_rbind(tib |> filter(.idx %in% entries) |> pull(data)),
```
dajmcdon, line 386: I prefer the "base R" pipe `|>` as well, but unless we depend on R (>=4.1), we can't use it.
```suggestion
      purrr::list_rbind(tib %>% filter(.idx %in% entries) %>% pull(data)),
```

dajmcdon (approved):

PR #451, opened 2025-03-21
feat: add step_/layer_ epi_YeoJohnson

dshemetov:
### Checklist

Please:

      dajmcdon.

- Resolves #{issue number}


air.toml
```diff
@@ -0,0 +1,8 @@
+[format]
```
dajmcdon, line 1: should probably add this to `.gitignore`? Specific to your setup?

R/layer_yeo_johnson.R
```diff
@@ -0,0 +1,249 @@
+#' Unormalizing transformation
+#'
+#' Will undo a step_epi_YeoJohnson transformation.
+#'
+#' @param frosting a `frosting` postprocessor. The layer will be added to the
+#'   sequence of operations for this frosting.
+#' @param lambdas Internal. A data frame of lambda values to be used for
```
dajmcdon, line 7: 1. This should likely not be labeled as internal. In principle, we'd like layers to operate outside a workflow. So allowing a user to pass the necessary object here should be allowed.

2. I don't like the name `lambdas`. Better to choose something non-jargon, but understandable. See e.g., https://www.tidymodels.org/learn/develop/models/#step-1.-register-the-model-modes-and-arguments

R/layer_yeo_johnson.R
```diff
@@ -0,0 +1,249 @@
...
+#'   inverting the transformation.
+#' @param ... One or more selector functions to scale variables
+#'   for this step. See [recipes::selections()] for more details.
+#' @param by A (possibly named) character vector of variables to join by.
```
dajmcdon, line 11: Is the intention a `join_by()` object? Even if not, it would be helpful to expand here along the lines of `?dplyr::left_join`.

dajmcdon: Alternatively, `@inheritParams step_population_scaling`. Actually, that should cover most of the examples.

R/layer_yeo_johnson.R
```diff
@@ -0,0 +1,249 @@
...
+  layer("epi_YeoJohnson", lambdas = lambdas, by = by, terms = terms, id = id)
+}
+
+#' @export
+#' @importFrom workflows extract_preprocessor
+slather.layer_epi_YeoJohnson <- function(object, components, workflow, new_data, ...) {
+  rlang::check_dots_empty()
+
+  # Get the lambdas from the layer or from the workflow.
+  lambdas <- object$lambdas %||% get_lambdas_in_layer(workflow)
```
dajmcdon, line 68: Should error here if we don't have a workflow.

R/layer_yeo_johnson.R
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    for (i in seq_along(col_names)) {
+      col <- col_names[i]
+      lambda_col <- paste0(".lambda_", original_outcome_cols[i])
+      components$predictions <- components$predictions %>%
+        rowwise() %>%
+        mutate(!!sym(col) := yj_inverse(!!sym(col), !!sym(lambda_col)))
+    }
+  }
```
dajmcdon, line 171: This part is all pretty dangerous.
1. `all_outcomes()` is a recipe selector. So you can use that, as long as you look in the `workflow`. 
2. Alternatively, even without the selector passed in `...`, you can look there to see which vars have the role `"outcome"`. 
3. But our typical case would want to apply the transformation to `starts_with(.pred)` to bring in the quantiles as well as the point prediction. I suspect that would hit your final `else` case but produce an error.
4. I'm not sure that we have an example with multiple aheads in separate columns, but if we did, presumably they'd have the same YJ parameters for different aheads of the same variable? This would be detectable from the workflow, which would be much safer than trying to get good behaviour with the string processing.

R/layer_yeo_johnson.R
```diff
@@ -0,0 +1,249 @@
...
+  this_recipe <- hardhat::extract_recipe(workflow)
+  if (!(this_recipe %>% recipes::detect_step("epi_YeoJohnson"))) {
+    cli_abort("`layer_epi_YeoJohnson` requires `step_epi_YeoJohnson` in the recipe.", call = rlang::caller_env())
+  }
+  for (step in this_recipe$steps) {
+    if (inherits(step, "step_epi_YeoJohnson")) {
+      lambdas <- step$lambdas
+      break
+    }
+  }
```
dajmcdon, line 247: this will be buggy if the step is used multiple times. Also, we want to make sure that the step applies to the outcome rather than the predictors (or other roles).

R/layer_yeo_johnson.R
```diff
@@ -0,0 +1,249 @@
...
+  is_neg <- ind_neg[["is"]]
+
+  nn_inv_trans <- function(x, lambda) {
+    if (abs(lambda) < eps) {
+      # log(x + 1)
+      exp(x) - 1
+    } else {
+      # ((x + 1)^lambda - 1) / lambda
+      (lambda * x + 1)^(1 / lambda) - 1
+    }
```
dajmcdon, line 214: To vectorize (can do the same below):
```suggestion
    out <- double(length(x))
    sm_lambdas <- abs(lambda) < eps
    out[sm_lambdas] <- exp(x[sm_lambdas]) - 1
    x <- x[!sm_lambdas]
    lambda <- lambda[!sm_lambdas]
    out[!sm_lambdas] <- (lambda * x + 1)^(1 / lambda) - 1
    out
  }
```

R/step_yeo_johnson.R
```diff
@@ -0,0 +1,397 @@
...
+#' @param limits A length 2 numeric vector defining the range to
+#'  compute the transformation parameter lambda.
+#' @param num_unique An integer where data that have fewer than this
+#'  many unique values will not be evaluated for a transformation.
+#' @param na_rm A logical indicating whether missing values should be
+#'  removed.
+#' @param skip A logical. Should the step be skipped when the recipe is
+#'  baked by [bake()]. On the `training` data, the step will always be
+#'  conducted (even if `skip = TRUE`).
+#' @param id A unique identifier for the step
```
dajmcdon, line 28: Can likely inherit from other steps to avoid duplication.

dajmcdon (commented):
Looking good so far. I've gotten through the layer stuff, but haven't really started on the step. There are a number of tasks we can discuss if I haven't provided enough detail.

air.toml
dshemetov: It's pretty early, but these are some formatting settings that match our styler settings. (If you haven't seen, the [air formatter](https://www.tidyverse.org/blog/2025/02/air/) is pretty new, but likely to be a common future formatter in R. Logan and I are both using it.)

R/layer_yeo_johnson.R
dshemetov: `yj_parameters`?

R/layer_yeo_johnson.R
dshemetov: Didn't know that was possible. I don't see a pattern of is_workflow(workflow) in other `slather.x` functions, so should I just do an inherits("workflow") check here or something?

dajmcdon: I suppose this is really for future extensions. Currently we don't have this functionality. So let's say, nice to have, but fine to comment and defer to future PR.

R/layer_yeo_johnson.R
dshemetov: Let me give this another shot, using `workflow` and assuming the typical case is `starts_with(.pred)`.

R/layer_yeo_johnson.R
dshemetov: Hm yea. I see a few ways through this:
- require that only a single step_yj is allowed; this would mean users have to bundle all their transformation requests in one step and all the transformations have to use the same hyperparameters `na_lambda_fill, limits, num_unique, na_rm`; that doesn't seem like a big limitation to me
- do a kind-of parentheses matching approach: the layer will need to count the yj layers prior to it and count backwards in the list of steps to find the step that matches it
  - doable, not super complex, and the general scheme might be handy for future layers, but the feature gain doesn't seem worth the effort right now

dajmcdon: Just for logging:
1. To access the processed recipe, use `hardhat::extract_recipe(workflow)`.
2. Inside is `last_term_info`. This tibble contains the names of all variables that ever existed, along with their roles.

Alternatively:
1. Access the mold with `hardhat::extract_mold(workflow)`.
2. all `y` are in a tibble at `$outcomes`

dajmcdon: The second is technically more robust within a workflow, if the preprocessor was other than a recipe (like a formula, for example).

R/step_yeo_johnson.R
```diff
@@ -0,0 +1,397 @@
...
+  # If it is an epi_df, check that the keys match.
+  # Imitating the pattern in step_adjust_latency().
+  if (!inherits(new_data, "epi_df") || is.null(attr(new_data, "metadata")$as_of)) {
+    new_data <- as_epi_df(
+      new_data,
+      as_of = object$forecast_date,
+      other_keys = object$metadata$other_keys %||% character()
+    )
+    attr(new_data, "metadata") <- object$metadata
+  }
```
dshemetov, line 198: So I removed this guardrail and I ran into an instance where `new_data` is a tibble in the tests. Seems out of scope to try and fix that, but wanted to mark that this is actually doing work right now.

Issue #262, opened 2023-10-26
[Discussion] A general framework for handling multi-ahead forecasters


dshemetov:
This includes:
- iterative models 
- smooth quantile reg

PR #452, opened 2025-03-21
using check_enough_train_data in practice

### LINKED ISSUE #333: Provide better error/recovery in residual quantiles ("zero-length inputs [...]")
Same message as #128.  I haven't fully tracked down the source, but it looks like maybe point predictions are missing here, so the stack trace / `options(error = recover())` may not be capturing the core cause.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(tidyr)
library(purrr)
library(ggplot2)
library(epidatr)
#> ! epidatr cache is being used (set env var EPIDATR_USE_CACHE=FALSE if not
#>   intended).
#> ℹ The cache directory is ~/.cache/R/epidatr.
#> ℹ The cache will be cleared after 14 days and will be pruned if it exceeds 4096
#>   MB.
#> ℹ The log of cache transactions is stored at ~/.cache/R/epidatr/logfile.txt.
library(epiprocess)
#> 
#> 
#>     filter
library(epipredict)
#> 
#> Attaching package: 'epipredict'
#> The following object is masked from 'package:ggplot2':
#> 
#>     layer

flusurv_analysis_issue <- as.Date("2019-08-01") %>%
  MMWRweek::MMWRweek() %>%
  {.$MMWRyear * 100L + .$MMWRweek}

flusurv_issue_data <-
  pub_flusurv(
    locations = "network_all",
    issues = epirange(123401, flusurv_analysis_issue)
  )
#> Warning: Loading from the cache at /home/fullname/.cache/R/epidatr; see
#> ~/.cache/R/epidatr/logfile.txt for more details.
#> This warning is displayed once every 8 hours.

flusurv_archive <- flusurv_issue_data %>%
  select(geo_value = location,
         time_value = epiweek,
         version = release_date,
         starts_with("rate_")) %>%
  as_epi_archive(compactify = TRUE)

archive <- flusurv_archive

forecast_dates <- seq(min(archive$DT$version) + 120L, archive$versions_end,
                      by = "6 weeks")

horizons <- c(0, 7, 14, 21, 28) # relative to forecast_date

example_forecaster <- function(snapshot_edf, forecast_date) {
  shared_reporting_latency <- as.integer(forecast_date - max(snapshot_edf$time_value))
  horizons %>%
    map(function(horizon) {
      snapshot_edf %>%
        arx_forecaster(
          outcome = "rate_overall",
          predictors = "rate_overall",
          args_list = arx_args_list(
            # (this is incomplete; latency often varies signficantly by covariate and can't be ignored, so we also need lag adjustment.)
            ahead = shared_reporting_latency + horizon,
            ## ahead = horizon,
            quantile_levels = c(0.1, 0.5, 0.9)
          )) %>%
        .$predictions %>%
        mutate(forecast_date = forecast_date,
               target_date = forecast_date + horizon)
    }) %>%
    bind_rows()
}

latest_edf <- archive %>% epix_as_of(.$versions_end)

unfaithful_forecasts <- latest_edf %>%
  # pretend we get observations about today, on today, with no revisions
  mutate(version = time_value) %>%
  as_epi_archive(versions_end = max(forecast_dates)) %>%
  epix_slide(
    # pretend version releases are on forecast dates
    ref_time_values = forecast_dates,
    before = 365000L, # 1000-year time window --> don't filter out any `time_value`s
    ~ example_forecaster(.x, .ref_time_value),
    names_sep = NULL
  ) %>%
  select(-time_value)
#> Warning in max.default(structure(numeric(0), class = "Date"), na.rm = FALSE):
#> no non-missing arguments to max; returning -Inf
#> Error in `map()` at rlang/R/dots.R:91:3:
#> ℹ In index: 1.
#> Caused by error in `lm.fit()`:
#> ! 0 (non-NA) cases
#> Backtrace:
#>      ▆
#>   1. ├─... %>% select(-time_value)
#>   2. ├─dplyr::select(., -time_value)
#>   3. ├─epiprocess::epix_slide(...) at dplyr/R/select.R:54:3
#>   4. │ └─x$slide(...)
#>   5. │   ├─... %>% ungroup()
#>   6. │   └─self$group_by()$slide(...) at dplyr/R/group-by.R:153:3
#>   7. │     └─base::lapply(...)
#>   8. │       └─epiprocess (local) FUN(X[[i]], ...)
#>   9. │         ├─dplyr::group_modify(...)
#>  10. │         ├─epiprocess:::group_modify.epi_df(...) at dplyr/R/group-map.R:156:3
#>  11. │         │ └─dplyr::dplyr_reconstruct(NextMethod(), .data)
#>  12. │         │   └─dplyr:::dplyr_new_data_frame(data) at dplyr/R/generics.R:196:3
#>  13. │         │     ├─row.names %||% .row_names_info(x, type = 0L) at dplyr/R/utils.R:18:3
#>  14. │         │     └─base::.row_names_info(x, type = 0L) at dplyr/R/utils.R:18:3
#>  15. │         ├─base::NextMethod()
#>  16. │         └─dplyr:::group_modify.data.frame(...)
#>  17. │           └─epiprocess (local) .f(.data, group_keys(.data), ...) at dplyr/R/group-map.R:166:3
#>  18. │             └─f(.data_group, .group_key, ref_time_value, ...)
#>  19. │               └─global example_forecaster(.x, .ref_time_value)
#>  20. │                 └─... %>% bind_rows()
#>  21. ├─dplyr::ungroup(.)
#>  22. ├─dplyr::bind_rows(.)
#>  23. │ └─rlang::list2(...) at dplyr/R/bind-rows.R:31:3
#>  24. ├─purrr::map(...) at rlang/R/dots.R:91:3
#>  25. │ └─purrr:::map_("list", .x, .f, ..., .progress = .progress) at purrr/R/map.R:129:3
#>  26. │   ├─purrr:::with_indexed_errors(...) at purrr/R/map.R:174:3
#>  27. │   │ └─base::withCallingHandlers(...) at purrr/R/map.R:201:3
#>  28. │   ├─purrr:::call_with_cleanup(...) at purrr/R/map.R:174:3
#>  29. │   └─.f(.x[[i]], ...)
#>  30. │     └─... %>% ...
#>  31. ├─dplyr::mutate(...)
#>  32. ├─epipredict::arx_forecaster(...) at dplyr/R/mutate.R:146:3
#>  33. │ ├─generics::fit(wf, epi_data)
#>  34. │ ├─epipredict:::fit.epi_workflow(wf, epi_data)
#>  35. │ ├─base::NextMethod()
#>  36. │ └─workflows:::fit.workflow(wf, epi_data)
#>  37. │   └─workflows::.fit_model(workflow, control)
#>  38. │     ├─generics::fit(action_model, workflow = workflow, control = control)
#>  39. │     └─workflows:::fit.action_model(...)
#>  40. │       └─workflows:::fit_from_xy(spec, mold, case_weights, control_parsnip)
#>  41. │         ├─generics::fit_xy(...)
#>  42. │         └─parsnip::fit_xy.model_spec(...)
#>  43. │           └─parsnip:::xy_form(...)
#>  44. │             └─parsnip:::form_form(...)
#>  45. │               └─parsnip:::eval_mod(...)
#>  46. │                 └─rlang::eval_tidy(e, env = envir, ...)
#>  47. ├─stats::lm(formula = ..y ~ ., data = data) at rlang/R/eval-tidy.R:121:3
#>  48. │ └─stats::lm.fit(...)
#>  49. │   └─base::stop("0 (non-NA) cases")
#>  50. └─base::.handleSimpleError(...)
#>  51.   └─purrr (local) h(simpleError(msg, call))
#>  52.     └─cli::cli_abort(...) at purrr/R/map.R:215:9
#>  53.       └─rlang::abort(...) at cli/R/rlang.R:45:3
```

<sup>Created on 2024-05-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

dsweber2:
### Checklist

Please:

- [X] Request a review from one of the current epipredict main reviewers:
      dajmcdon.

I was going through the pile of issues and #333 was a good test case for `check_enough_train_data`, so I added it to `arx_forecaster()`. To get it working, I also needed to modify `check_enough_train_data` by replicating the logic in the bake step, since it should be applied most importantly to the test data, and not the train data.

- Resolves #333


dshemetov:
A test time data check seems reasonable! However:
- don't quite think the `n` should be the same at prep and bake time
  - for one they seem like they should be separate parameters
  - I'm also concerned the `n` in the bake step case will tend to always just be 1 (since check_enough_train will come after all predictor construction) ... 
  - also feels like this is treading on similar ground that get_oversized_test_data() refactors 
- the fn name should probably remove `train` then 

For instance, in the test you added
```r
test_that("warns if there's not enough data to predict", {
  edf <- tibble(
    geo_value = "ct",
    time_value = seq(as.Date("2020-10-01"), as.Date("2023-05-31"), by = "day"),
  ) %>%
    mutate(value = seq_len(nrow(.)) + rnorm(nrow(.))) %>%
    # Oct to May (flu season, ish) only:
    filter(!between(as.POSIXlt(time_value)$mon + 1L, 6L, 9L)) %>%
    # and actually, pretend we're around mid-October 2022:
    filter(time_value <= as.Date("2022-10-12")) %>%
    as_epi_df(as_of = as.Date("2022-10-12"))
  edf %>% filter(time_value > "2022-08-01")

  debugonce(bake.check_enough_train_data)
  expect_error(
    edf %>% arx_forecaster("value"),
    class = "epipredict__not_enough_train_data"
  )
})

Browse[1]> non_na_data %>%
               group_by(across(all_of(.env$object$epi_keys)))
# A tibble: 444 x 7
   geo_value time_value value lag_0_value lag_7_value lag_14_value ahead_7_value
   <chr>     <date>     <dbl>       <dbl>       <dbl>        <dbl>         <dbl>
 1 ct        2020-10-15  16.4        16.4        7.39        0.565          20.1
 2 ct        2020-10-16  15.9        15.9        9.05        2.86           23.5
 3 ct        2020-10-17  17.6        17.6        8.91        3.24           23.7
 4 ct        2020-10-18  18.2        18.2        8.62        3.11           24.3
 5 ct        2020-10-19  18.3        18.3       13.8         6.69           27.0
 6 ct        2020-10-20  20.1        20.1       13.3         6.16           27.1
 7 ct        2020-10-21  21.0        21.0       16.6         7.52           29.0
 8 ct        2020-10-22  20.1        20.1       16.4         7.39           29.0
 9 ct        2020-10-23  23.5        23.5       15.9         9.05           29.7
10 ct        2020-10-24  23.7        23.7       17.6         8.91           32.4

Browse[1]> object$n
[1] 3
```

Where n=3 is coming from the number of predictors we have (which is a default in the case of unspecified n), but we certainly can do a prediction here with only one data point.

R/arx_forecaster.R
```diff
@@ -171,7 +171,9 @@ arx_fcast_epi_workflow <- function(
     step_epi_ahead(!!outcome, ahead = args_list$ahead)
   r <- r %>%
     step_epi_naomit() %>%
-    step_training_window(n_recent = args_list$n_training)
+    step_training_window(n_recent = args_list$n_training) %>%
+    check_enough_train_data(all_predictors(), n = args_list$check_enough_data_n, skip = FALSE)
+
   if (!is.null(args_list$check_enough_data_n)) {
```
dshemetov, line 177: How come this is separate from the `check_enough_train_data` below?

dsweber2: the short of it is I didn't read closely enough to notice that that was included.

dsweber2:
> for one they seem like they should be separate parameters

Good point.

> I'm also concerned the n in the bake step case will tend to always just be 1 (since check_enough_train will come after all predictor construction) ...

I was wondering about the multi-ahead case, but I think that too should only need one point to do the prediction.
> also feels like this is treading on similar ground that get_oversized_test_data() refactors

That is happening before any steps are applied, so it can't really make this kind of check. For example, `get_test_data` can't tell you which columns don't have enough data.


I think I'll switch the default back to `skip = TRUE`, and adjust the usage in `arx_forecaster` so that it checks for a fixed `n` (fixed to 1, unless we can think of other cases).

We may just want to change the name to `check_enough_data` because it was basically flexible enough to cover train or test before I made this PR. The changes I made narrowed behavior so that it only drops `NA`'s from the specified columns to check, and enables bake-time running if `skip = FALSE`, which are both things we'd want in either case.

dshemetov:
> I was wondering about the multi-ahead case
FWIW there are barriers before we can support [multi-ahead forecasters](https://github.com/cmu-delphi/epipredict/issues/262).
> That is happening before any steps are applied, so it can't really make this kind of check. For example, get_test_data can't tell you which columns don't have enough data.
Right, but wasn't the gist of our refactor plan basically to push the filtering to after the predict step? So then new_data would first pass through hardhat::forge -> recipes::bake, where it possibly receive the very basic "is there at least 1 data point per group" data check that doesn't seem essential, since the same effective check can phrased as "is there at least prediction per group".
> We may just want to change the name to check_enough_data because it was basically flexible enough to cover train or test before I made this PR. The changes I made narrowed behavior so that it only drops NA's from the specified columns to check, and enables bake-time running if skip = FALSE, which are both things we'd want in either case.
Changing name sounds good to me.

dsweber2:
> since the same effective check can phrased as "is there at least prediction per group".

Looking at #333, the problem is coming up in the middle of the prediction step, with completely unrelated errors coming through before we ever get the output from `predict`, so I think having a step that catches it just before `fit` would be useful

R/check_enough_data.R
```diff
@@ -0,0 +1,195 @@
...
+          all_of(.env$col_names),
+          ~ sum(!is.na(.x)) < .env$step_obj$n
+        ),
+        .groups = "drop"
+      ) %>%
+      summarise(across(all_of(.env$col_names), any), .groups = "drop") %>%
+      unlist() %>%
+      names(.)[.]
+
+    if (length(cols_not_enough_data) == 0) {
```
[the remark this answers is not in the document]
dsweber2: yeah it's a bit weird, but it's intentional. `cols_not_enough_data` is checking each  column separately, rather than all together. This covers the case where no single column is missing data

dshemetov (approved):
lgtm and I double checked that it correctly gives a helpful error for Logan's example in #333

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+  res$id <- x$id
+  res$n <- x$n
+  res$epi_keys <- x$epi_keys
+  res$drop_na <- x$drop_na
+  res
+}
+
+check_enough_data_core <- function(epi_df, step_obj, col_names, train_or_predict) {
+  epi_df <- epi_df %>%
+    group_by(across(all_of(.env$step_obj$epi_keys)))
```
dajmcdon, line 143: Here and below, do you need the `.env` prefix?

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+  if (step_obj$drop_na) {
+    any_missing_data <- epi_df %>%
+      mutate(any_are_na = rowSums(across(any_of(.env$col_names), ~ is.na(.x))) > 0) %>%
+      # count the number of rows where they're all not na
+      summarise(sum(any_are_na == 0) < .env$step_obj$n, .groups = "drop")
+    any_missing_data <- any_missing_data %>%
+      summarize(across(all_of(setdiff(names(any_missing_data), step_obj$epi_keys)), any)) %>%
+      any()
+
+    # figuring out which individual columns (if any) are to blame for this darth
```
dajmcdon, line 153:
```suggestion
    # figuring out which individual columns (if any) are to blame for this dearth
```

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+    }
+  } else {
+    # if we're not dropping na values, just count
+    cols_not_enough_data <- epi_df %>%
+      summarise(
+        across(
+          all_of(.env$col_names),
+          ~ dplyr::n() < .env$step_obj$n
+        )
+      )
```
dajmcdon, line 184:
```suggestion
      summarise(across(all_of(.env$col_names), ~ dplyr::n() < .env$step_obj$n))
```

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+#'
+#' `check_enough_data` creates a *specification* of a recipe
+#'  operation that will check if variables contain enough data.
+#'
+#' @param recipe A recipe object. The check will be added to the
+#'  sequence of operations for this recipe.
+#' @param ... One or more selector functions to choose variables for this check.
+#'  See [selections()] for more details. You will usually want to use
+#'  [recipes::all_predictors()] and/or [recipes::all_outcomes()] here.
+#' @param n The minimum number of data points required for training. If this is
```
dajmcdon, line 11: Agree with @dshemetov from elsewhere: can we rename this arg? Perhaps `min_required_n`?

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+#' @param epi_keys A character vector of column names on which to group the data
+#'   and check threshold within each group. Useful if your forecaster trains
+#'   per group (for example, per geo_value).
+#' @param drop_na A logical for whether to count NA values as valid rows.
+#' @param role Not used by this check since no new variables are
+#'  created.
+#' @param trained A logical for whether the selectors in `...`
+#' have been resolved by [prep()].
+#' @param columns An internal argument that tracks which columns are evaluated
+#'   for this check. Should not be used by the user.
```
dajmcdon, line 22: Let's remove this arg here. It can be added / used in the `*_new()` constructor.

(Core `{recipes}` exposes args that the user shouldn't and wouldn't want to change. My preference is not to have them unless explicitly mentioned as important (`role` would be an example here)).

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+#' @export
+prep.check_enough_data <- function(x, training, info = NULL, ...) {
+  col_names <- recipes::recipes_eval_select(x$terms, training, info)
+  if (is.null(x$n)) {
+    x$n <- length(col_names)
+  }
+
+  check_enough_data_core(training, x, col_names, "train")
```
dajmcdon, line 97: Is this needed at prep time? Is it enough to only have it process during bake? Prep runs only during model training. Bake runs at train time _and_ at test time (unless `skip = TRUE`).

R/epi_workflow.R
```diff
@@ -270,6 +270,6 @@ forecast.epi_workflow <- function(object, ..., n_recent = NULL, forecast_date =
     hardhat::extract_preprocessor(object),
     object$original_data
   )
-
+  test_data
```
dajmcdon, line 273: extraneous?
```suggestion
```

R/check_enough_data.R
```diff
@@ -0,0 +1,201 @@
...
+      summarise(across(all_of(.env$col_names), any), .groups = "drop") %>%
+      unlist() %>%
+      # Select the names of the columns that are TRUE
+      names(.)[.]
+  }
+
+  if (any_missing_data) {
+    cli_abort(
+      "The following columns don't have enough data to {train_or_predict}: {cols_not_enough_data}.",
+      class = "epipredict__not_enough_data"
```
dajmcdon, line 198: I don't really mind attaching the class everywhere, but is it still helpful?

dajmcdon (changes requested):

R/check_enough_data.R
dsweber2: It makes catching specific errors possible if you want to be able to bypass it and it's better for tests.

R/check_enough_data.R
dsweber2: I was adopting something that Dmitry had used in this step originally; I believe the point was to make sure there's never confusion with the column names. Probably not a problem in this case, though with the original `step_obj` being just `x` it was much more likely to have a collision.

R/check_enough_data.R
dsweber2: So the one advantage of having it in both is that it allows us to print `The following columns don't have enough data to train: x and y.` if it breaks during prep and `The following columns don't have enough data to predict: x and y.` if it breaks during prediction. It ends up only running once either way, and the tests all pass with it removed from `prep`.

R/check_enough_data.R
dajmcdon: ... maybe `min_observations`?

R/check_enough_data.R
dajmcdon: I'm fine to leave it, but I'm not convinced it does what you want. Is there a corner case in which it can pass at prep time but fail at bake time? In that case, model training would error, but the message would say something about predicting.

dajmcdon (approved):

R/check_enough_data.R
dsweber2: I guess one *could* substitute a different dataset between baking and prepping, but substituting a dataset with less data seems unlikely.

PR #458, opened 2025-04-08
extend to quantile_dist, exclude multi-output

dsweber2:

This grew out of addressing Dan's concerns about the way that outcome handling is done in `layer_Yeo_Johnson`. I opted to default to `starts_with(".pred")` or whichever columns the user specifies. I dropped support for multiple output columns with different yj_parameters (aheads would be included as different rows rather than columns), and normalize all of them using whichever column corresponds to the output in the recipe's `term_info`

In addition, I noticed that the inverse yeo_johnson transform wasn't working on `quantile_dist` objects. So I rewrote it so that it would, and rewrote the forward transform to match the format.

To add a check for that I added `vec_arith.quantile_pred.quantile_pred`, though there are ways around it that wouldn't be too nasty.


dsweber2:
checks+formatting passed locally, I guess they're not running b/c I'm not targeting dev

dshemetov (approved):
Overall looks good, thanks for this! I made a few tweaks
- added and sharpened a few tests after checking code coverage
- fixed some quantile_pred arithmetic edge cases (note: these were needed for tests where we difference quantile pred objects, but might be handy in general)

PR #463, opened 2025-04-23
Get test data rework

dsweber2:
### Checklist

Please:

- [ ] Request a review from one of the current epipredict main reviewers:
      dajmcdon.
- [ ] Make sure to bump the version number in `DESCRIPTION` and `NEWS.md`.
- [ ] Describe changes made in NEWS.md, making sure breaking changes

make `get_test_data` retrieve a year by default, and make predict narrow down to the forecast_date

- Resolves #{issue number}

### DIFF
--- a/R/arx_forecaster.R
+++ b/R/arx_forecaster.R
@@ -47,20 +47,19 @@ arx_forecaster <- function(
   if (!is_regression(trainer)) {
     cli_abort("`trainer` must be a {.pkg parsnip} model of mode 'regression'.")
   }
-
   wf <- arx_fcast_epi_workflow(epi_data, outcome, predictors, trainer, args_list)
   wf <- fit(wf, epi_data)
 
   # get the forecast date for the forecast function
   if (args_list$adjust_latency == "none") {
-    forecast_date_default <- max(epi_data$time_value)
+    reference_date_default <- max(epi_data$time_value)
   } else {
-    forecast_date_default <- attributes(epi_data)$metadata$as_of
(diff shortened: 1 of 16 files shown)


dajmcdon:
@dsweber2 How important do you think this methodology is? Just a thought, but what if we 
1. deprecate this function completely;
2. only call `forecast()` in examples (rather than `predict()`);
3. Inside of `forecast()`, just predict the entire training data, then filter to the reference date.

dsweber2:
if the fitting is expensive, it's probably best to predict as few days as possible (I've run into this with grf). Having something that handles that filtering is probably worthwhile.

That said, the most important part of this PR is definitely the filtering the output in the way `predict` runs so that it's only outputting forecasts for the right days. In practice that was the part that tripped me up more when avoiding `get_test_data` in exploration-tooling

dajmcdon:
Just to be clear, `get_test_data()` has nothing to do with model fitting, only the amount of preprocessing necessary. Computation could hit in two ways:
1. Long running steps (`step_growth_rate()` would be such) that bog down on excess data; although the growth rate calculation is an example of a case where the current implementation becomes very fragile, and this fix would help;
2. Actually predicting on the inputs; this is unlikely to cost very much in most cases. Tree predictions are very fast, as are linear models.

Where are you imagining the hit is occurring? I agree that point 1 is a reason to avoid training on excess data (and one reason for implementing this function in the first place), but my current thinking, after much use, is that the fragility is a larger issue than the computation. Do you have examples where the cost to 1 is burdensome?

Issue #468, opened 2025-07-25
Bug: get_test_data uses na.omit() on all data table columns, rather than just predictors/outcomes


dshemetov:
The line in question is [here](https://github.com/cmu-delphi/epipredict/blob/dev/R/get_test_data.R#L60-L63).

I realize we have consensus that `get_test_data()` will be re-written, but 
- it's been that way for over a year now, so a temp fix might still be worth it here
- if a temp fix is not worth it, figured I'd report this so we don't re-create this bug in the new implementation

(Sorry this reprex isn't very minified)

```r
library(epipredict)
library(epiprocess)
library(epidatr)
library(data.table)
library(dplyr)
library(tidyr)
library(ggplot2)
library(magrittr)
library(purrr)

#Fetch data from Delphi Epidata API 
nhsn <- pub_covidcast(
  source = "nhsn",
  signals = "confirmed_admissions_covid_ew_prelim",
  geo_type = "state",
  time_type = "week",
  geo_values = "*",
  time_values = epirange(202001, 202601),
  issues = epirange(202001, 202601)
) |>
  # The version date column is called `issue` in the Epidata API. Rename it.
  rename(version = issue, confirmed_covid_adm = value) |>
  as_epi_archive(compactify = TRUE)

nhsn_original <- nhsn
nhsn <- nhsn_original
#Plot of monthly snapshots of the data
geo_choose <- "ca"

forecast_dates <- as.Date("2025-03-09")

covid_adm_data <- bind_rows(
  # Snapshotted data for the version-faithful forecasts
  map(
    forecast_dates,
    ~ nhsn |>
      epix_as_of(.x) |>
      mutate(version = .x)
  ) |>
    bind_rows() |>
    mutate(version_faithful = "Version faithful"),
  # Latest data for the version-un-faithful forecasts
  nhsn |>
    epix_as_of(nhsn$versions_end) |>
    mutate(version_faithful = "Version un-faithful")
)

# Enter debug mode here
debug(epipredict::get_test_data)
archive_cases_dv_subset_faux %>%
  epix_as_of(as.Date("2025-03-09")) %>%
  arx_forecaster("confirmed_covid_adm", "confirmed_covid_adm", args_list = arx_args_list(
          ahead = 7,
          lags = c(0, 7, 14, 21),
          adjust_latency = "extend_ahead"
        ))

```

Issue #267, opened 2023-11-04
`get_test_data` returning `NA`'s if there are `NA`'s in the most recent data


dsweber2:
This seems like a bug. An example of what I mean:

```
    geo_values <-jhu$geo_value %>% unique()
    one_day_nas <- tibble(
      geo_value = geo_values,
      time_value = as.Date("2022-01-01"),
      case_rate = NA,
      death_rate = runif(length(geo_values))
    )
    second_day_nas <- one_day_nas %>%
      mutate(time_value = as.Date("2022-01-02"))
    jhu_nad <- jhu %>%
      as_tibble() %>%
      bind_rows(one_day_nas, second_day_nas) %>%
      as_epi_df()
    attributes(jhu_nad)$metadata$as_of <- max(jhu_nad$time_value) + 3
```
The example workflow is unfortunately buried in the guts of exploration tooling; `arx_forecaster` actually does do the right thing.

dsweber2:
kind of related to #106

dsweber2:
@dshemetov how would this interact with the work you've been doing with `forecast` and `get_test_data`? Should we hold off on this until that's done?

dshemetov:
I'm hoping that the work there ends up resolving this.

dsweber2:
Mostly deprecated in light of #293

dshemetov:
IIRC this was addressed via a generous use of clear_last_minute_nas, different datasets (weekly), and the introduction of our own get_oversized_test_data() function. The reprex above is worth tracking, to make sure it's fixed by get_test_data() in #293 though.

PR #477, opened 2026-05-06
fix: display latency adjustment info on canned forecasters (#447)

### LINKED ISSUE #447: `print.canned_pred` checks contents of non-existent `pre` (partial-matched to unintended element)
https://github.com/cmu-delphi/epipredict/blob/7f08d409d60677d4007cdbaa6bf272b46e035e6e/R/canned-epipred.R#L115
```r
print.canned_epipred <- function(x, name, ...) {
[...]
  if ("actions" %in% names(x$pre) && "recipe" %in% names(x$pre$actions)) {
```

Presumably, `x$pre` is supposed to be the preprocessor, but it actually partial-matches to `x$predictions`.

musing: if not for this coincidence, `warnPartialMatchDollar` wouldn't have even caught this and we'd just be testing vs. `NULL`.  This is also an issue for `epi_archive`s.  Is there any sort of list subclass that makes `$` do exact matches and produce an error rather than `NULL` when the requested key has no match?

gregfaletto:
Resolves #447.

## Summary

`print.canned_epipred()` was supposed to surface latency-adjustment information (e.g., `Aheads adjusted per column: case_rate=2, death_rate=3`) for canned forecasters whose recipe contains a `step_adjust_latency`. It never actually did, because the gating conditional checked for a top-level `pre` field on the canned forecaster object — but canned forecasters' top-level fields are `predictions`, `epi_workflow`, and `metadata`. The conditional was always FALSE, and the entire latency-info block was dead code.

The fix replaces the conditional with `!is.null(x$epi_workflow$pre$actions$recipe)`, which correctly tests for a recipe preprocessor on the embedded workflow. The latency-info display now works for all four recipe-based canned forecasters (`arx_forecaster`, `arx_classifier`, `cdc_baseline_forecaster`, `flatline_forecaster`); `climatological_forecaster` correctly remains unaffected (it overwrites `pre` with a `mold`-only list and has no `step_adjust_latency`).

## Test coverage

The existing snapshot tests in `tests/testthat/test-snapshots.R` already exercised `arx_forecaster` with `adjust_latency = "extend_lags"` (`out2`) and `"extend_ahead"` (`out3`) but never observed the latency line because of the bug. Their snapshots regenerate to include the new lines (one `+` line each in `_snaps/snapshots.md` — purely additive). Two supplementary `expect_match("adjusted", ...)` assertions are added next to those snapshot calls as defensive guards: if a future contributor mechanically `snapshot_accept`s after an unrelated upstream change drops the "adjusted" line, the explicit assertions still catch the regression.

## Bookkeeping

- `NEWS.md`: new `# epipredict 0.2.3` section with one bullet.
- `DESCRIPTION`: version bumped to 0.2.3.
- `.gitignore` and `.Rbuildignore`: 3 lines each adding `.workflow/`, `.plans/`, and `.claude/` to keep personal tooling directories out of git tracking and out of the package build tarball. Happy to split into a separate PR if you prefer.

## Out of scope

The original issue mentions a broader concern about partial-matching vulnerabilities in `epi_archive` / `epi_df` list subclasses. Those types live in `epiprocess`, not `epipredict`. Within this repo, I confirmed via grep that `R/canned-epipred.R:115` was the only `x$pre` access where `x` was not a workflow object; the others (`R/epi_recipe.R:308`, `R/extract.R:88`, `R/model-methods.R:89`) all operate on workflow objects where `pre` is the actual field name. A broader audit belongs as a follow-up `epiprocess` issue.

## Test plan

- `devtools::test()` — 910 passing, 0 failing, 2 skips. The skips (`test-get_test_data.R:47` "NA fill behaves as desired" and `:85` "forecast date behaves") are pre-existing on `dev`, added in commit `09fbfd812` (July 2024) as placeholders during the locf-into-`step_adjust_ahead` refactor; the underlying `get_test_data` rework is tracked by PR #463 and issues #468 / #267.
- `devtools::check(document = FALSE, vignettes = FALSE)` — 0 errors. 1 environmental warning (`parsnip` binary built under R 4.5.2 vs local R 4.5.0 — won't fire in CI). 2 pre-existing notes on `dev` (top-level `Makefile`/`pkgdown-watch.R`, deep `:::` import list).
- `lintr::lint_package()` — zero new findings on lines added by this PR.
- Manually verified before/after: `arx_forecaster` with `step_adjust_latency` now shows the latency-adjustment bullet at the bottom of `print()` output.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -22,3 +22,6 @@
 ^.lintr$
 ^.venv$
 ^inst/templates$
+^\.workflow$
+^\.plans$
+^\.claude$
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ inst/doc
 renv.lock
 renv/
 .Renviron
+.workflow/
+.plans/
+.claude/
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: epipredict
 Title: Basic epidemiology forecasting methods
-Version: 0.2.2
+Version: 0.2.3
 Authors@R: c(
     person("Daniel J.", "McDonald", , "daniel@stat.ubc.ca", role = c("aut", "cre")),
     person("Ryan", "Tibshirani", , "ryantibs@cmu.edu", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,10 @@
 
 Pre-1.0.0 numbering scheme: 0.x will indicate releases, while 0.0.x will indicate PR's.
 
+# epipredict 0.2.3
+
+- Fix `print.canned_epipred()` so the latency-adjustment information actually displays for canned forecasters that include `step_adjust_latency` in their recipe (#447).
+
 # epipredict 0.2.2
 
 - Fix `autoplot.epi_workflow()` to correctly handle the response variable and avoid errors related to `.response`.
--- a/R/canned-epipred.R
+++ b/R/canned-epipred.R
@@ -112,7 +112,7 @@ print.canned_epipred <- function(x, name, ...) {
     "At forecast date{?s}: {.val {fds}},",
     "For target date{?s}: {.val {tds}},"
   ))
-  if ("pre" %in% names(x) && "actions" %in% names(x$pre) && "recipe" %in% names(x$pre$actions)) {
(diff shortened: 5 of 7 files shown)


dsweber2 (approved):
Thanks for the fix! The summary is... kind of overkill and made it a bit harder to follow that the key improvement was including `epi_workflow` in the path.

